PR #14, opened 2018-02-01
Move skylight to dev_dep and check whether loaded or not

minsikzzang:
* Move `skylight` to development_dependency
* Only use `skylight` helpers when it's loaded


grossadamm:
We need to connect with a tool such as TravisCI to make sure tests pass automatically for new PRs. I love this PR and will pull it down to test manually. I'm all for merging once confirmed.

minsikzzang:
@grossadamm cool, I create new PR to hook up TravisCI, if you want :)

waynehoover:
Don't you need to move it out of the Gemfile section too? As juwelier generates the .gemspec correct?

minsikzzang:
@waynehoover yes, you're right, I will move it to dev in Gemfile, too

minsikzzang:
@waynehoover instead of changing in `Gemfile`, added `gemspec` so we can use only one source of dependencies

minsikzzang:
@grossadamm I think the specs need to be updated. They shouldn't heavily rely on performance comparison 🙏

grossadamm:
I think a balance needs to be struck between performance testing as well as flexibility to adding new features. They pass locally, but we need to make sure that travis passes reliably as well.

grossadamm:
They pass locally for me, as well as another coworker. I'm going to merge this and we can figure out how to balance the TravisCI builds after :)

PR #33, opened 2018-02-02
Get rid of temp variables

apneadiving:
Here is some refactoring mostly to get rid of temporary variables. Hopefully it helps making it even faster ⚡


shishirmk:
Thank you for make the code more concise. Do you mind creating a the pull request against dev?

Issue #24, opened 2018-02-01
Possibly skewing of benchmark results


ocha:
I noticed 2 issues with how benchmarks are set up.
1) `build_ams_movies` builds owner, movie_type and actor objects while `build_movies` only sets up ids and builds objects just in time when serializing the movies array (and not building owner object at all). It looks like AMS serialization will use more memory and trigger more GC (be slower) because of the way it is set up. Both test setups should initialize same amout of objects/object types
2) AMS is not set up to use OJ even though it supports it. comparing performance of `serializable_hash` of both gems or using OJ with both gems should be fairer comparison. 

could these 2 issue cause some problems with "fairness" of the benchmark?

minsikzzang:
I have a PR https://github.com/Netflix/fast_jsonapi/pull/36 for enabling OJ to AMS.

PR #36, opened 2018-02-02
Enable OJ and optimize performance test

minsikzzang:
* As mentioned in https://github.com/Netflix/fast_jsonapi/issues/24, it enables OJ for AM.
* Disable GC in performance benchmark


minsikzzang:
Oh By the way, I based on master, because master was latest, I can move it back to dev, if neded

shishirmk:
Thank you for adding this. It makes it explicit that OJ is being used for both AMS and fast_jsonapi during the JSON string comparison.

PR #38, opened 2018-02-03
Test against Ruby 2.2

nicolasleger:


guilleiguaran:
FYI Ruby 2.2 only receive security updates and will be EOL the next month.

PR #42, opened 2018-02-03
Add Primalize gem for performance comparison

jgaskins:
Primalize has different opinions about structure than the JSON API, so the output isn't the same, but I thought it might be useful to have another performance datapoint.

I just noticed I left some debugging output in there. I'll remove it if you're interested in merging.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -97,7 +97,9 @@ GEM
       multi_xml (~> 0.5)
       rack (>= 1.2, < 3)
     oj (3.4.0)
-    primalize (0.3.4)
+    primalize (0.3.5)
+    primalize-jsonapi (0.1.2)
+      primalize
     public_suffix (3.0.1)
     rack (2.0.3)
     rack-test (0.8.2)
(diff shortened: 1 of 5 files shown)


grossadamm:
@jgaskins right now we are focused on just the json api spec. It is an interesting comparison, but I'm not sure if it makes sense to have right now. If we open this gem up to be more flexible, it would definitely make sense.

jgaskins:
No worries, this is good to know. I wasn't sure if this gem was more focused on the "fast" part or the "JSONAPI" part. :-)

If there were a JSONAPI adapter for Primalize, would that be a more helpful performance datapoint?

jgaskins:
I had some spare time to hammer out a JSONAPI adapter for Primalize. This gem is still quite a bit faster, but it appears both are miles ahead of AMS. From my local test runs:

> Serialize to JSON string 1000 with includes and meta

| Serializer | Records | Time |
|----------|----------|------|
| AMS serializer | 1000 | 314.78 |
| Fast serializer | 1000 | 12.78 |
| Primalize | 1000 | 25.21 |

shishirmk:
I think we have so many comparisons already. I am not sure if more comparisons will add any value. So I am going to close this. Thank you for the pull request

PR #43, opened 2018-02-04
Refactoring

smakagon:
Hi, I've refactored a little bit of `object_serializer` module.

I've renamed `default_record_type` to `reflected_record_type` because we use reflection to guess that type. Also, I've added memoization to that method, so we don't calculate type every time we call `reflected_record_type`.

I've extracted logic of initialization into a separate private method.

Besides that, I've combined `@record` and `@records` into one variable `@resource` and added `is_collection?` helper that should help to figure out if it's a collection.

There are couple smaller changes.
I used existing performance specs and `benchmark-ips` gem to verify that performance didn't change after refactoring.

Thanks for a great gem!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,8 +26,7 @@ pkg
 #*.tmproj
 #tmtags
 
-# For vim:
-#*.swp
+*.swp
 
 # For redcar:
 #.redcar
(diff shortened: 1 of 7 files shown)


shishirmk:
Love the changes you have made here @smakagon. Thank you for diving in and cleaning things up. I think we are going to have some merge conflicts with #45. We can figure out how to deal with them.

smakagon:
Thanks, @shishirmk!
I'll be more than happy to resolve any merge conflicts if you decide to merge this PR.

Issue #41, opened 2018-02-03
Support for CamelCase attributes


pratikpandit05:
We're looking at json serializers specifically for serializing a PORO. 

One of the core requirements for serialization is that the snake_cased attribute keys be transformed into CamelCased ones to support the json standards being followed by our organization and also some others like Google, w3resource, schema.org, etc. 

Will such a request be considered for this gem? It can be a special configuration item. If yes, I am open to putting a tech design of sorts

shishirmk:
@pratikpandit05 checkout the pull request. That adds the flexibility to choose any kind of key transform including CamelCase and camelCaseLower

pratikpandit05:
+1. Thank you.

MarcusRiemer:
As the mentioned PR #45 has been merged: Is there any ETA for the version containing the change to be released?

shishirmk:
@MarcusRiemer We are planning to release a 1.1.0 version of the gem soon (in a week or two). We are figuring out if we should release it with custom attributes feature being discussed in pull requests or just the things we have merged so far are enough.

iSunRise:
@shishirmk Hey, thank you for the awesome gem. I tried this feature #45 and think that you should mention in the readme that `set_key_transform` method must be called before `attributes` otherwise it won't work. 
I added `set_key_transform` to the bottom of my serializer and it took some time to realize why it doesn't work.

shishirmk:
Released as a part of version 1.1.0

PR #45, opened 2018-02-04
Feature/key transform

shishirmk:
This pull request the adds the key_transform functionality available in AMS. You can transform the keys and record types using 4 transforms. camel, camel_lower, underscore, dash.

This should address the question/feature request in #41 

Additionally this is not backward compatible to use_hyphen. So we will have to bump up the version number accordingly.


NullVoxPopuli:
I have a gem (that AMS(, and maybe others? Use)) called case_transform. It also has a native extension counterpart called case_transform-rust-extensions, which reduce conversion time by ~20x depending on the transform.

shishirmk:
@NullVoxPopuli This is happening during class initialization so it shouldn't affect the serialization time. We are avoiding gems with native extensions for now. Here is a pull request where i have only two gems as runtime dependencies #40

README.md
```diff
@@ -134,6 +135,26 @@ json_string = MovieSerializer.new(movie).serialized_json
...
+  set_key_transform :camel
+end
+```
+Here are examples of how these options transform the keys
+
+```ruby
+set_key_transform :camel # "some_key" => "SomeKey"
+set_key_transform :camel_lower # "some_key" => "someKey"
+set_key_transform :dash # "some_key" => "some-key"
+set_key_transform :underscore # "some_key" => "some_key"
```
bf4: fwiw, the fastest key transform in AMS is `:unaltered` :)

PR #46, opened 2018-02-04
Use Object#public_send to respect attributes visibility

guilleiguaran:
While working on other feature I found that fast_jsonapi serializers are able to call private methods in the record object and I think this isn't the intended behavior.

This replaces `Object#send` with`Object#public_send` to avoid the calling of private methods in serializers.


shishirmk:
Thank you for fixing that

Issue #35, opened 2018-02-02
has_one serialization fails for nil association 


elrosa:
Attempt to serialize model with `has_one` association fails if association is nil:

**`NoMethodError (undefined method 'id' for nil:NilClass)`**

`(fast_jsonapi-1.0.16/lib/extensions/has_one.rb:12)`

If associated model exists, it gets serialized correctly.

Example:
```Ruby
class User < ApplicationRecord
  has_one :profile
end
 
class Profile < ApplicationRecord
   belongs_to :user
end

class UserSerializer
  include FastJsonapi::ObjectSerializer
  attributes :id, :name
  has_one :profile
end

UserSerializer.new(User.create(name: "John")).serialized_json
NoMethodError (undefined method 'id' for nil:NilClass)

```

grossadamm:
Great find! We'll get a test to replicate and fix.

grossadamm:
Looks like @jeremyjung is on this. Thanks!

christophersansone:
@shishirmk We have a merged PR... should this issue be closed?

shishirmk:
Fixed in Version 1.1.0

PR #48, opened 2018-02-05
Allow has_one nil association

jeremyjung:
Attempt to fix issue #35 .

Any feedback or suggestions on where to add a test would be appreciated.  I created a sample rails app to verify functionality.


grossadamm:
Thanks @jeremyjung! If you could, could you add some tests? For guidance, I would suggest at least adding a test to `object_serializer_spec`

jeremyjung:
The existing movie context (Movie, Actor, etc.) doesn’t have any has_one relations in it.  I wasn’t sure if I should modify the existing models in there or create new models specifically to test has_one.

shishirmk:
@jeremyjung @grossadamm This issue could be fixed by doing a nil check in line 12 of this file https://github.com/Netflix/fast_jsonapi/blob/master/lib/extensions/has_one.rb

Something like this

```ruby
association(:#{name}).reader&.id
```

jeremyjung:
@shishirmk Good call with the dig method.  I'm having a little trouble figuring out the appropriate tests for this.

Putting this test in `object_serializer_spec.rb` doesn't interact with ActiveRecord and so it never touches the has_one extension method.
```ruby
    it 'returns correct json when has_one returns nil' do
      supplier.account_id = nil
      json = SupplierSerializer.new(supplier).serialized_json
      serializable_hash = JSON.parse(json)
      expect(serializable_hash['data']['relationships']['account']['data']).to be nil
    end
```

Another possibility could be just testing the extension method in the `active_record_spec.rb` to make sure it doesn't return nil:
```ruby
    it 'has account_id method return nil if not present' do
      expect(Supplier.find(@supplier_id_without_account).account_id).to eq nil
    end
```

In addition, any thoughts on removing the has_one extension method completely?  Without the extension method I think we could still get similar behavior by using:
```ruby
    public_send(:object_method_name)&.id
```

shishirmk:
@jeremyjung I think the test can go to `active_record_spec.rb`. The test you have in the comment looks good. 

Throughout the gem we try to support serialization with just the `id_method_name` so that we dont have to create an object unnecessarily when we call `object_method_name`. It is also really helpful when we just have the id in our database and the record is in another micro service. In such cases we can serialize the record by avoiding a network call. Hope that explains why we have chosen to use this approach.

shishirmk:
The &. Operator is available only ruby 2.3 onwards. That's why the builds are failing on 2.2.* version. 

Do you mind using something more verbose?

shishirmk:
@jeremyjung Thank you for making all the necessary changes

Issue #5, opened 2018-02-01
Custom attributes


pplant:
Will it be possible to add custom attributes to serializers that will delegate to a function within the serializer class. E.g.:

    attributes :fullname
    
    def fullname
        "#{object.firstname} #{object.lastname}"
    end

Are you guys willing to accept a pull request implementing such a feature?

NielsKSchjoedt:
+1

ansonhoyt:
I use this feature of AMS regularly, as an API-specific attribute often doesn't belong in my model.

saturnflyer:
I was curious about this and I think some code like this could allow you to get around the missing custom attributes without changes to the gem:

```ruby
require 'delegate'
class CustomSerializer
  include FastJsonapi::ObjectSerializer

  attributes :fullname

  def initialize(resource, options={})
    super
    @record = CustomAttributes.new(@record)
  end

  class CustomAttributes < DelegateClass(YourRecordClass)
    def fullname
      "#{firstname} #{lastname}"
    end
  end
end
```

One could simplify it if it's a common enough pattern

sriniwasgr:
@pplant Appreciate your input. I believe we have this functionality for relationships if can use a class method (say: in a decorator) and use this on the serializer.

Example:

`has_many :locations, object_method_name: :places`

You could create a decorator with the method 'places' and include it on the model. In your case, i see you need that for an attribute, we will be happy to provide this functionality. Feel free to create a PR to dev if you are interested in getting it done. Please make sure it is performant. :)

TrevorHinesley:
Not sure if this is implied in this request, but we also override existing attributes to format them differently for JSON responses.

grossadamm:
This makes a ton of sense to have an easy way to handle this.

shishirmk:
@TrevorHinesley @pplant the workaround is to add the method to the model instead of the serializer.

If you would like to define it in the serializer that is not possible today. Happy to look at pull request if you get around to it.

TrevorHinesley:
@shishirmk unfortunately that isn’t a viable option for our case. For instance, we would have to use a different name than the column obviously, which means the attribute in the response would be named differently than the column. But if we want the attribute to have the same name as the column, that wouldn’t be possible without overriding its getter. We don’t want to override the getter since it’s just for JSON serialization. 

I don’t think I have the availability right now to work on a PR, but if someone submits one that would be awesome! Thank you all for the awesome library.

[...]

PR #49, opened 2018-02-05
Add support for custom attribute methods.

guilleiguaran:
This adds support for custom attribute methods as suggested in #5.

```ruby
class MovieSerializer
  attributes :name, :release_year, :title_with_year

  def title_with_year
    "#{record.name} (#{record.release_year})"
  end
end
```

This required substantial changes in the API, moving the SerializationCore methods from class to instance level making it more similar to AMS API.

### DIFF
--- a/lib/fast_jsonapi/object_serializer.rb
+++ b/lib/fast_jsonapi/object_serializer.rb
@@ -50,8 +50,8 @@ module FastJsonapi
 
       return serializable_hash unless @resource
 
-      serializable_hash[:data] = self.class.record_hash(@resource)
-      serializable_hash[:included] = self.class.get_included_records(@resource, @includes, @known_included_objects) if @includes.present?
+      serializable_hash[:data] = self.record_hash
+      serializable_hash[:included] = self.get_included_records(@includes, @known_included_objects) if @includes.present?
       serializable_hash
     end
 
@@ -61,8 +61,9 @@ module FastJsonapi
       data = []
       included = []
       @resource.each do |record|
-        data << self.class.record_hash(record)
-        included.concat self.class.get_included_records(record, @includes, @known_included_objects) if @includes.present?
+        serializer = self.class.new(record)
+        data << serializer.record_hash
+        included.concat serializer.get_included_records(@includes, @known_included_objects) if @includes.present?
       end
 
       serializable_hash[:data] = data
@@ -72,7 +73,7 @@ module FastJsonapi
     end
 
     def serialized_json
-      self.class.to_json(serializable_hash)
+      self.to_json(serializable_hash)
     end
 
     private
--- a/lib/fast_jsonapi/serialization_core.rb
+++ b/lib/fast_jsonapi/serialization_core.rb
@@ -16,76 +16,78 @@ module FastJsonapi
       end
     end
 
-    class_methods do
-      def id_hash(id, record_type)
-        return { id: id.to_s, type: record_type } if id.present?
-      end
+    def record
+      @resource
+    end
 
-      def ids_hash(ids, record_type)
-        return ids.map { |id| id_hash(id, record_type) } if ids.respond_to? :map
-        id_hash(ids, record_type) # ids variable is just a single id here
-      end
+    def id_hash(id, record_type)
+      return { id: id.to_s, type: record_type } if id.present?
+    end
 
-      def attributes_hash(record)
-        attributes_to_serialize.each_with_object({}) do |(key, method_name), attr_hash|
-          attr_hash[key] = record.public_send(method_name)
+    def ids_hash(ids, record_type)
+      return ids.map { |id| id_hash(id, record_type) } if ids.respond_to? :map
+      id_hash(ids, record_type) # ids variable is just a single id here
+    end
+
+    def attributes_hash
+      self.class.attributes_to_serialize.each_with_object({}) do |(key, method_name), attr_hash|
+        attr_hash[key] = respond_to?(method_name) ? public_send(method_name) : record.public_send(method_name)
+      end
+    end
+
+    def relationships_hash(relationships = nil)
+      relationships = self.class.relationships_to_serialize if relationships.nil?
+
+      relationships.each_with_object({}) do |(_k, relationship), hash|
+        name = relationship[:key]
+        id_method_name = relationship[:id_method_name]
+        record_type = relationship[:record_type]
+        empty_case = relationship[:relationship_type] == :has_many ? [] : nil
+        hash[name] = {
+          data: ids_hash(record.public_send(id_method_name), record_type) || empty_case
+        }
+      end
+    end
+
+    def record_hash
+      if self.class.cached
+        record_hash = Rails.cache.fetch(record.cache_key, expires_in: self.class.cache_length) do
+          temp_hash = id_hash(record.id, self.class.record_type) || { id: nil, type: self.class.record_type }
+          temp_hash[:attributes] = attributes_hash if self.class.attributes_to_serialize.present?
+          temp_hash[:relationships] = {}
+          temp_hash[:relationships] = relationships_hash(self.class.cachable_relationships_to_serialize) if self.class.cachable_relationships_to_serialize.present?
+          temp_hash
         end
+        record_hash[:relationships] = record_hash[:relationships].merge(relationships_hash(self.class.uncachable_relationships_to_serialize)) if self.class.uncachable_relationships_to_serialize.present?
+        record_hash
+      else
+        record_hash = id_hash(record.id, self.class.record_type) || { id: nil, type: self.class.record_type }
+        record_hash[:attributes] = attributes_hash if self.class.attributes_to_serialize.present?
+        record_hash[:relationships] = relationships_hash if self.class.relationships_to_serialize.present?
+        record_hash
       end
+    end
 
-      def relationships_hash(record, relationships = nil)
-        relationships = relationships_to_serialize if relationships.nil?
+    def to_json(payload)
+      MultiJson.dump(payload) if payload.present?
+    end
 
-        relationships.each_with_object({}) do |(_k, relationship), hash|
-          name = relationship[:key]
-          id_method_name = relationship[:id_method_name]
-          record_type = relationship[:record_type]
-          empty_case = relationship[:relationship_type] == :has_many ? [] : nil
-          hash[name] = {
-            data: ids_hash(record.public_send(id_method_name), record_type) || empty_case
-          }
-        end
-      end
+    # includes handler
 
-      def record_hash(record)
-        if cached
-          record_hash = Rails.cache.fetch(record.cache_key, expires_in: cache_length) do
-            temp_hash = id_hash(record.id, record_type) || { id: nil, type: record_type }
-            temp_hash[:attributes] = attributes_hash(record) if attributes_to_serialize.present?
-            temp_hash[:relationships] = {}
-            temp_hash[:relationships] = relationships_hash(record, cachable_relationships_to_serialize) if cachable_relationships_to_serialize.present?
-            temp_hash
-          end
-          record_hash[:relationships] = record_hash[:relationships].merge(relationships_hash(record, uncachable_relationships_to_serialize)) if uncachable_relationships_to_serialize.present?
-          record_hash
-        else
-          record_hash = id_hash(record.id, record_type) || { id: nil, type: record_type }
-          record_hash[:attributes] = attributes_hash(record) if attributes_to_serialize.present?
-          record_hash[:relationships] = relationships_hash(record) if relationships_to_serialize.present?
-          record_hash
-        end
-      end
-
-      def to_json(payload)
-        MultiJson.dump(payload) if payload.present?
-      end
-
-      # includes handler
-
-      def get_included_records(record, includes_list, known_included_objects)
-        includes_list.each_with_object([]) do |item, included_records|
-          object_method_name = @relationships_to_serialize[item][:object_method_name]
-          record_type = @relationships_to_serialize[item][:record_type]
-          serializer = @relationships_to_serialize[item][:serializer].to_s.constantize
-          relationship_type = @relationships_to_serialize[item][:relationship_type]
-          included_objects = record.send(object_method_name)
-          next if included_objects.blank?
-          included_objects = [included_objects] unless relationship_type == :has_many
-          included_objects.each do |inc_obj|
-            code = "#{record_type}_#{inc_obj.id}"
-            next if known_included_objects.key?(code)
-            known_included_objects[code] = inc_obj
-            included_records << serializer.record_hash(inc_obj)
-          end
+    def get_included_records(includes_list, known_included_objects)
+      includes_list.each_with_object([]) do |item, included_records|
+        object_method_name = self.class.relationships_to_serialize[item][:object_method_name]
+        record_type = self.class.relationships_to_serialize[item][:record_type]
+        serializer = self.class.relationships_to_serialize[item][:serializer].to_s.constantize
+        relationship_type = self.class.relationships_to_serialize[item][:relationship_type]
+        included_objects = record.send(object_method_name)
+        next if included_objects.blank?
+        included_objects = [included_objects] unless relationship_type == :has_many
+        included_objects.each do |inc_obj|
+          code = "#{record_type}_#{inc_obj.id}"
+          next if known_included_objects.key?(code)
+          known_included_objects[code] = inc_obj
+          included_records << serializer.new(inc_obj).record_hash
         end
       end
     end
--- a/spec/lib/object_serializer_with_custom_method_spec.rb
+++ b/spec/lib/object_serializer_with_custom_method_spec.rb
@@ -1,0 +1,18 @@
+require 'spec_helper'
+
+describe FastJsonapi::ObjectSerializer do
+  include_context 'movie class'
+
+  context 'when including custom methods defined in serializer in attributes' do
+    it 'returns correct hash when serializable_hash is called' do
+      serializable_hash = MovieSerializerWithCustomMethod.new([movie, movie]).serializable_hash
+      expect(serializable_hash[:data].length).to eq 2
+      expect(serializable_hash[:data][0][:attributes].length).to eq 3
+      expect(serializable_hash[:data][0][:attributes]).to have_key(:title_with_year)
+
+      serializable_hash = MovieSerializerWithCustomMethod.new(movie_struct).serializable_hash
+      expect(serializable_hash[:data][:attributes].length).to eq 3
+      expect(serializable_hash[:data][:attributes]).to have_key(:title_with_year)
+    end
+  end
+end
--- a/spec/lib/serialization_core_spec.rb
+++ b/spec/lib/serialization_core_spec.rb
@@ -7,29 +7,29 @@ describe FastJsonapi::ObjectSerializer do
     it 'returns correct hash when id_hash is called' do
       inputs = [{id: 23, record_type: :movie}, {id: 'x', record_type: 'person'}]
       inputs.each do |hash|
-        result_hash = MovieSerializer.send(:id_hash, hash[:id], hash[:record_type])
+        result_hash = MovieSerializer.new(movie).send(:id_hash, hash[:id], hash[:record_type])
         expect(result_hash[:id]).to eq hash[:id].to_s
         expect(result_hash[:type]).to eq hash[:record_type]
       end
 
-      result_hash = MovieSerializer.send(:id_hash, nil, 'movie')
+      result_hash = MovieSerializer.new(movie).send(:id_hash, nil, 'movie')
       expect(result_hash).to be nil
     end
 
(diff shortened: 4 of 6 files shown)


guilleiguaran:
An alternative that requires fewer changes and is more performant can be implemented with this API:

```ruby
class MovieSerializer
  attributes :name, :release_year, :title_with_year

  def self.title_with_year(record)
    "#{record.name} (#{record.release_year})"
  end
end
```

lib/fast_jsonapi/object_serializer.rb
```diff
@@ -61,8 +61,9 @@ def hash_for_collection
       data = []
       included = []
       @resource.each do |record|
-        data << self.class.record_hash(record)
-        included.concat self.class.get_included_records(record, @includes, @known_included_objects) if @includes.present?
+        serializer = self.class.new(record)
```
saturnflyer: how does initializing an object with every iteration affect performance on large collections?

guilleiguaran: Sharing numbers in current dev vs this branch:

dev:

```
Serialize to JSON string 250 records
AMS serializer  250        83.96 ms
Fast serializer 250        3.4 ms

Serialize to Ruby Hash 250 records
AMS serializer  250        77.04 ms
Fast serializer 250        2.73 ms

Serialize to JSON string 1000 records
AMS serializer  1000       303.68 ms
Fast serializer 1000       9.94 ms

Serialize to Ruby Hash 1000 records
AMS serializer  1000       300.06 ms
Fast serializer 1000       12.39 ms
```

this branch:

```
Serialize to JSON string 250 records
AMS serializer  250        78.39 ms
Fast serializer 250        3.71 ms

Serialize to Ruby Hash 250 records
AMS serializer  250        75.98 ms
Fast serializer 250        2.74 ms

Serialize to JSON string 1000 records
AMS serializer  1000       307.92 ms
Fast serializer 1000       10.63 ms

Serialize to Ruby Hash 1000 records
AMS serializer  1000       279.43 ms
Fast serializer 1000       8.98 ms
```

guilleiguaran: More benchmarks (with includes and meta):

dev:

```
Serialize to JSON string 250 with includes and meta
AMS serializer  250        112.38 ms
Fast serializer 250        4.62 ms

Serialize to Ruby Hash 250 with includes and meta
AMS serializer  250        101.24 ms
Fast serializer 250        4.35 ms

Serialize to JSON string 1000 with includes and meta
AMS serializer  1000       422.2 ms
Fast serializer 1000       20.34 ms

Serialize to Ruby Hash 1000 with includes and meta
AMS serializer  1000       394.01 ms
Fast serializer 1000       18.07 ms
```

this branch:

```
Serialize to JSON string 250 with includes and meta
AMS serializer  250        104.63 ms
Fast serializer 250        4.96 ms

Serialize to Ruby Hash 250 with includes and meta
AMS serializer  250        101.09 ms
Fast serializer 250        4.3 ms

Serialize to JSON string 1000 with includes and meta
AMS serializer  1000       401.42 ms
Fast serializer 1000       21.19 ms

Serialize to Ruby Hash 1000 with includes and meta
AMS serializer  1000       389.07 ms
Fast serializer 1000       21.17 ms
```

christophersansone:
@guilleiguaran Thanks for taking a crack at this.  I was about to build an implementation myself!

I wonder if this can be a lot simpler to implement and potentially faster to render if we just used a block or Proc.  I was thinking of the following:

```ruby
class MovieSerializer

  attribute :title_with_year do |record|
    "#{record.name} (#{record.release_year})"
  end

end
```

Then the serializer can just store the block and execute it with each record iteration.

christophersansone:
As someone who uses AMS a lot, I am not only jumping for joy at the speed boosts that this project promises, but also how it can better handle memory usage as the size and number of documents grow (see https://github.com/rails-api/active_model_serializers/issues/2051).  I have not done any memory benchmarking between the two, but a quick code review seems to indicate that memory consumption will not be an issue due to the lack of object instantiation.

I would be cautious about an implementation that requires instantiation of a new serializer object for each record.

guilleiguaran:
@christophersansone thanks for your feedback, I understand your concern and that was the reason I proposed another alternative myself in https://github.com/Netflix/fast_jsonapi/pull/49#issuecomment-363122269

I have added the `get_process_mem` gem in my local copy of the code to measure the change in memory between the `dev` and this branch in every `it` block of the `context 'when testing performance of serialization'` block, the result is this:

dev branch:

```
Memory: 91.0859375 MB
Memory: 91.125 MB
Memory: 91.125 MB
Memory: 91.125 MB
```

this branch:

```
Memory: 91.08984375 MB
Memory: 91.12890625 MB
Memory: 91.12890625 MB
Memory: 91.12890625 MB
```

The difference appears to be `3KB` when instantiating 4000 objects.

That said, I'll be happy with whatever implementation the authors choose.

christophersansone:
@guilleiguaran Thanks for providing some actual data around my assertion about memory concerns.  Yes, 3KB does not seem like much, but I would imagine that the growth would be fairly linear, until the GC kicks in.  It could be a problem in high traffic situations.  And my understanding of Ruby internals is weak, but it is possible that as more functionality is added to the serializer code base, the more memory each object instantiation will take.

The reality is that it takes time, effort, and memory to instantiate and clean up Ruby objects, and if there is an opportunity to design an API that avoids instantiation without sacrificing code readability, it is worth serious consideration... especially for a gem whose name starts with the word `fast`. :smile:

lib/fast_jsonapi/object_serializer.rb
grossadamm: There is some cost, and incidentally, the performance cost using new instances on collections is something we actively debated about during development. On my machine the results are slightly different, and while the speed is impacted, it's still "performant" IMHO.

I do like the idea of @christophersansone's approach though. And I'm wondering if we can combine that approach with the instance method definition. The class could inspect for instance methods and convert them to procs using ruby goodness. Then, serialization would not require new instances and we could maintain the instance method syntax.

This is something I have not explored, and it may add more complexity than it warrants. It may also add some level confusion for developers on this project since an instance method would not be acting as an instance method... 

For giggles, here is a comparison in branches bumped to 10,000 records. I don't know why you would serialize 10,000 records following the JSON API spec, so it's a pretty contrived example, but it does illustrate the cost a little better.

dev:
```
Serialize to JSON string 10000 with includes and meta
AMS serializer  10000      4791.46 ms
Fast serializer 10000      234.48 ms

Serialize to Ruby Hash 10000 with includes and meta
AMS serializer  10000      4484.9 ms
Fast serializer 10000      193.3 ms
```

this branch:
```
Serialize to JSON string 10000 with includes and meta
AMS serializer  10000      4741.36 ms
Fast serializer 10000      360.68 ms

Serialize to Ruby Hash 10000 with includes and meta
AMS serializer  10000      4529.02 ms
Fast serializer 10000      214.24 ms
```

grossadamm: I also wonder if it's better to take a performance hit by merging this in, and then iterating to a more performant approach.

christophersansone: > It may also add some level confusion for developers on this project since an instance method would not be acting as an instance method

+1 on that.  We would definitely have to take into account attribute methods that call other methods (e.g. if you have a complex method to calculate an attribute that calls other methods).  A developer would expect that to just work, so just cherry-picking certain instance methods could be trouble.  And using instance variables would not behave as expected, either.  The serializer could potentially have some assertions for the developer to help avoid some common pitfalls.  But if we can avoid all of this by just having a block with the attribute definition, that works.

A couple different options for those who would really want a separate method (though they would have to be class methods):

```ruby
attribute :title_with_year { |r| MovieSerializer.title_with_year(r) }

def self.title_with_year(record)
  ...
end
```

or

```ruby
attribute :title_with_year, class_method: :title_with_year

def self.title_with_year(record)
  ...
end
```

Or we just have a generator function (or recommend that they do it themselves).  Or we just provide the capability and offer some suggestions in the readme...

```ruby
def self.class_attribute(attr)
  attribute(attr) { |record| self.class.send(attr, record) }
end

class_attribute :title_with_year
```

Lots of options to avoid serializer instantiation.  My opinion would be to start with the simplest, most performant solution and then iterate on it based on feedback or limitations.

guilleiguaran: @christophersansone @grossadamm I like the simplicity of the Proc approach but in my opinion, there are many cases where it's convenient to break the calculation of attributes into multiple methods and I'm not sure if we can achieve that using Procs. e.g:

```ruby
class FeedItemSerializer
  # ignoring other attributes
  attributes :cooked_count, :likes_count, :liked_photo_comments_count

  # ignoring other associations
  has_many :liked_photo_comments

  def liked_photo_comments
    return [] unless triggered_by_like?
    photo_comments_liked_by_author.recent.limit(4)
  end

  def liked_photo_comments_count
    return unless triggered_by_like?
    photo_comments_liked_by_author.count
  end

  def cooked_count
    approved_photo_comments.count
  end

  def likes_count
    recipe_likes.count
  end

  private

    def triggered_by_like?
      # ...
    end

    def photo_comments_liked_by_author
      # ...
    end

    def recipe_likes
      # ...
    end

    def approved_photo_comments
      # ...
    end
end
```

guilleiguaran: Another case we rely on method attributes and goodies of OOP is in our fragment caching (used to cache attributes in serializers where isn't desirable to cache entire objects)

```ruby
# Using Rails.cache for the example, the logic of our Caching System is a bit more complex.
class ApplicationSerializer < ActiveModel::Serializer
  def self.expire_cache_fragment!(name)
    Rails.cache.delete("ams_fragment_cache_#{name}")
  end

  protected
  def cache_fragment(name)
    Rails.cache.fetch("ams_fragment_cache_#{name}") do
      yield
    end
  end
end
```
And then we can use the `cache_fragment` method in all the subclasses of ApplicationSerializer, e.g:

```ruby
class MovieSerializer < ApplicationSerializer
  attributes :background_actors

  def background_actors
    cache_fragment("background_actors") do
      # Heavy computation to serialize all background actors ...
    end
  end
end
```

christophersansone: @guilleiguaran @grossadamm Agreed that complex calculations could be cumbersome with the block syntax, and the AMS syntax is quite nice for doing this.  But is it worth the performance hit for every single serialization, regardless of whether this extra syntactic sugar is needed?  My opinion is that we should not force some performance penalty at all times, but potentially allow the developer to opt into it as needed.

One solution is to simply use class methods, passing in the record each time.  This example assumes attribute block syntax and nothing else, but we can design something better if the concept is a good one:

```ruby
attribute :title_with_year { |r| MovieSerializer.title_with_year(r) }

def self.title_with_year(record)
  ...
end
```

If that is not sufficient, then consider having a separate decorator class that can be instantiated.

```ruby
attribute :title_with_year { |r| MovieDecorator.new(r).title_with_year }
```

We could also implement the ability to define a decorator in the serializer, which would instantiate an instance of the decorator for each record, e.g.:

```ruby
class MovieSerializer
  decorator MovieDecorator
  attribute :title_with_year
end

decorator MovieDecorator
  include FastJsonapi::ObjectDecorator
  def title_with_year
    ...
  end
end
```

The decorator would basically define an `initialize` with the record, and it would implement `method_missing` to send all undefined methods to the record.

Lots of options here, just trying to come up with some ideas that work but don't impose any performance loss or unnecessary memory consumption by default.

shishirmk:
@guilleiguaran, @christophersansone Thank you for making pull requests to support this feature. In my view the best developer experience is the one AMS has for custom attributes. It's simple and easy to understand and it would be familiar syntax to everyone using this gem. So the best case scenario is if we can get the AMS developer experience with 0-10% performance degradation.

I am not sure if it's possible, but i want to give that a shot. Any ideas of how we can achieve that?

christophersansone:
@shishirmk Thanks!  My concern with the "10% performance budget" is that this project is so young, and decisions we make today that cause performance degradation could have an increasing impact as the code base grows.  I imagine AMS did not start out as slow as it is today, but design decisions made early on have made it virtually impossible to undo without causing breaking API changes.

There are three high-level possibilities here:

1. **Perform serialization at the class level rather than the instance level.**  Without some crazy Ruby trickery that I am unaware of, I don't think we can accomplish this while offering the same syntax as AMS.

2. **Perform serialization at the instance level, and create new serializer instances for each record, like AMS.**  My opinion is that this will incur a performance degradation that will increase over time and be hard to undo once the syntax is public.

3.  **Perform serialization at the instance level, and build a strategy to reuse instances.**  Some sort of instance pool would work here, where an instance is checked out at the start of serialization and then checked back in upon completion.  Attempting to check out from an empty pool can simply create a new instance, so that way there is always one available.  This may strike a good balance between performance while offering the preferred syntax.  The only added concern would be that state (i.e. instance variables) would be maintained across executions, so developers would have to be made aware of that.  We could offer some sort of method to reset the variables if this is a concern.

Another concern about instance-level serialization in general is the introduction of _naming collisions_ between attributes and the methods inside the serializer, e.g. `record`, `record_hash`, `id_hash`, etc.  Defining an attribute with a name that already exists would be problematic... at the very least, we would need the ability to reference a different method name.  Even if developers avoided the _current_ list of method names, we could introduce a new method on the serializer that now collides with an attribute being used in the wild.  My preference would be to design a solution that avoided naming collisions all together.

There are several options here, each with their own set of trade-offs.  I would like to propose one last idea that I believe strikes a nice balance:

```ruby
class MovieSerializer
  attributes :name, :release_year, :title_with_year

  module HelperMethods
    def title_with_year
      ...
    end
  end
end
```

Upon loading the class, it can detect if the `HelperMethods` module has been defined, and if so, adjust the serialization strategy to instantiate a "decorator" for each record, containing those methods.  This syntax is not exactly AMS but it is very close.  This solution has the following benefits:

1. Clear syntax, very similar to AMS.

2. Clear logical separation between the serializer and the helper methods.

3. Prevents naming collisions between the serializer and the helper methods.

4. Serialization still occurs at the class level.

5. Performance is only affected in the serializers that have `HelperMethods`.

6. We can still potentially provide the block syntax for attributes as well, in cases where the calculation is simple and / or the performance cost of `HelperMethods` needs to be avoided.

7. The (memory) footprint of the decorator instances will remain small and fixed, regardless of how much `FastJsonapi::ObjectSerializer` continues to grow in size and scope.

Several options here.  It's awesome to see that the community is rallying around this project so quickly, and that we are getting a wide set of opinions with this feature.

sriniwasgr:
@shishirmk I am with @christophersansone on the point about the 10% budget. We definitely should look to solve it without performance regression. This discussion is great, we should think about strategies we have not thought of yet.

guilleiguaran:
> Upon loading the class, it can detect if the HelperMethods module has been defined, and if so, adjust the serialization strategy to instantiate a "decorator" for each record, containing those methods. 

This gives me an idea, let's assume that we have this serializer:

```ruby
class MovieSerializer
  attributes :name, :release_year, :title_with_year

  def title_with_year
      # ...
  end
end
```

and copy/pasting most of your idea:

Upon loading the class, it can detect if instance methods have been defined, and if so, adjust the serialization strategy to instantiate a "decorator" for each record, containing those methods.

christophersansone:
@guilleiguaran I imagine it could do this sort of detection and select an optimal strategy, such as class-based serialization vs. instance-based serialization, or reusable instances vs. one-time-use instances. My understanding is that Ruby does not allow cherry-picking methods from one class and insert them into another (decorator) class, so I think this syntax would limit us to instantiating the actual serializer.  It is definitely an option.

shishirmk:
@guilleiguaran @christophersansone what if we used the serializer instance itself as the decorator for the record?

christophersansone:
@shishirmk Just to clarify, are you asking _what if we did the serialization at the class level but instantiated instances of the serializer solely for the purpose of "decorating"_?  It's an option.  One side effect would be that these decorator methods would have to be `public` since they are called externally, which would not be obvious to those coming from AMS.

guilleiguaran:
> @guilleiguaran @christophersansone what if we used the serializer instance itself as the decorator for the record?

@shishirmk @christophersansone I'll work in a proof of concept implementation of this now to see how it goes.

shishirmk:
@christophersansone @guilleiguaran We could move some of the class methods and make them instance methods. I dont think that would have a significant impact on performance. I think it's best if we don't create more than one instance of the serializer any class per call to `serializable_hash`.

christophersansone:
In general I would be wary of a solution that provides an AMS-like syntax... but breaks some canonical Ruby syntax, such as the use of instance variables, private and protected methods, etc.

Fundamentally, I believe whatever solution that is agreed upon should embrace the architectural differences between the two projects, even if it means a slightly different syntax.  _AMS is designed so that a serializer instance will serialize a single record_, so its attribute definition syntax can assume instance methods, an `object` instance variable, etc.  Here, _a single FastJsonapi serializer instance can serialize many records_, so there are some crucial differences are fundamentally incompatible with the AMS syntax.

I have another idea.  There are a lot of opinions flying around about what the right solution could be.  What if FastJsonapi were instead _less opinionated_ about how to implement this?  Instead, we could just provide some primitive hooks and some examples in the documentation.

For example:

```ruby
class MovieSerializer
  attributes :name, :release_year, :title_with_year

  # override to convert the record into something that accepts the defined attributes and relationships
  def self.serializable_object(record)
      # default implementation: "return record"
      MovieDecorator.new(record)
  end

  # override to change how a particular attribute is serialized
  def self.serialize_attribute(record, attribute_name, options)
    # default implementation: "return record.send(attribute_name)"
  end
end
```

This way, (a) it's completely customizable, and (b) we can better allow some good patterns to emerge with some real-world use.  And we can ensure that FastJsonapi remains fast out of the box.

guilleiguaran:
@christophersansone @shishirmk I've been exploring the option of using the serializer class itself as decorator, I've implemented it in this commit: https://github.com/guilleiguaran/fast_jsonapi/commit/7c10d5e17602bfce20bd6ecb2d4d928049ee4e08

I've done a comparison with 10,000 records (with fewer records the difference is insignificant) to illustrate the cost of the change:

dev branch:
```
Serialize to JSON string 10000 records
AMS serializer  10000      3158.22 ms
Fast serializer 10000      122.85 ms

Serialize to Ruby Hash 10000 records
AMS serializer  10000      3044.81 ms
Fast serializer 10000      92.41 ms
```

feature_custom_attr_serializer_v2 branch:
```
Serialize to JSON string 10000 records
AMS serializer  10000      3002.7 ms
Fast serializer 10000      128.84 ms

Serialize to Ruby Hash 10000 records
AMS serializer  10000      2855.9 ms
Fast serializer 10000      112.68 ms
```

wdyt?

guilleiguaran:
@christophersansone thank you very much for your review, I've applied your suggestions/fixes and now there isn't any difference in performance between `dev` and the branch.

Running again perf tests with 10,000 records:

dev branch:
```
Serialize to JSON string 10000 records
AMS serializer  10000      3026.65 ms
Fast serializer 10000      115.48 ms

Serialize to Ruby Hash 10000 records
AMS serializer  10000      2877.32 ms
Fast serializer 10000      88.09 ms
```

feature_custom_attr_serializer_v2 branch:
```
Serialize to JSON string 10000 records
AMS serializer  10000      3019.01 ms
Fast serializer 10000      114.63 ms

Serialize to Ruby Hash 10000 records
AMS serializer  10000      2916.93 ms
Fast serializer 10000      88.89 ms
```

I'll add perf tests for the custom attributes and open a PR later today if other folks in this thread are happy with this solution.

shishirmk:
@guilleiguaran @christophersansone You guys have been awesome. Thank you for trying out all the different ideas we came up with. I think custom attributes is a very important feature a lot of people have asked for it. I would like to have it in the upcoming 1.1.0 release which we are planning to release next week. Keeping in mind the time we have, performance and ease of use I am planning to merge in #54. We can continue to refine the feature after we release 1.1.0

TrevorHinesley:
Just want to say, this was incredible to read. Big brains here. Thanks for the feature!!

TrevorHinesley:
Is this ready to merge in? Looks like everything is passing other than some conflicts to fix.

axsuul:
Isn't this redundant now that https://github.com/Netflix/fast_jsonapi/pull/54 is merged in?

guilleiguaran:
@axsuul as @shishirmk #54 is merged but we will continue exploring the idea of supporting custom attributes through methods since blocks aren't suitable for all situations.

monarch0111:
@shishirmk @christophersansone @guilleiguaran I was thinking to add custom attribute method support using below snippet.

```ruby
%w(custom_attr_1 custom_attr_2).each do |custom_attr| 
    attribute custom_attr.to_sym do |object|
        self.send(custom_attr, object)
    end
end

def self.custom_attr1(object)
    # do something
end

def self.custom_attr2(object)
    # do something
end
```
What will be the performance issues while using this? Any advice would be appreciated.

shishirmk:
We have diverged too much so closing this

Issue #44, opened 2018-02-04
Benchmarks vs jsonapi-rb?


NullVoxPopuli:
Jsonapi-rb was written by @beauby, who also used to work on AMS (i did too). It's also way faster than AMS.

I saw the medium blog post, and i think more details about the benchmark woulld be good, too. 
For example, in my own benchmarks, with 200 records with 2 included relationships each,. I found that jsonapi-rb is 5 to 6 times faster than AMS. I can share the benchmarrk files with you if it'd help.

I have some doubts about the speed increase mentioned in the medium article, which is why i bring this up. 

I could do my own benchmarks, but i'm in the car right now. :)

Thanks for supporting/making-more-known the jsonapi spec. :D

shishirmk:
Thanks for reaching out. We are putting together a longer writeup. Our benchmarks can be run very easily If you just run rspec you will see the results. More information about how the benchmarks are setup take a look at this file 

https://github.com/Netflix/fast_jsonapi/blob/master/spec/lib/object_serializer_performance_spec.rb

We had a very specific goal in mind so we got the speed increase by sacrificing some flexibility, focusing just on JSON API and assuming homogenous lists.

ahx:
@shishirmk Thank you for sharing this and thinking about putting together a longer writeout. It would be great if you could add one or two sentences about *why* fast_jsonapi is faster than AMS. I am very interested in this sort of performance optimization, but I currently have no clue about where to start or what techniques to avoid in order to become/stay fast.

minsikzzang:
benchmark against `jsonapi-rb` added https://github.com/Netflix/fast_jsonapi/pull/52

sriniwasgr:
@NullVoxPopuli  @shishirmk  Here is the [link](https://github.com/Netflix/fast_jsonapi/blob/master/performance_methodology.md) to the performance doc. Hopefully it helps in answering your question.

amitsuryavanshi:
I did some benchmarking with [https://github.com/amitsuryavanshi/json-parser-benchmark](script) and results (Iterations per second) were,
Warming up --------------------------------------
          jsonapi-rb    10.000  i/100ms
Calculating -------------------------------------
          jsonapi-rb    111.401  (±20.6%) i/s -      1.580k in  15.058330s


Warming up --------------------------------------
        fastjson-api   118.000  i/100ms
Calculating -------------------------------------
        fastjson-api      1.270k (± 9.6%) i/s -     18.880k in  15.014813s

[...]

Issue #12, opened 2018-02-01
JSONAPI Suite Integration


richmolj:
I’m the creator of [JSONAPI Suite](https://jsonapi-suite.github.io/jsonapi_suite) which currently uses [jsonapi-rb](http://jsonapi-rb.org) for the serialization layer. Have you benchmarked against jsonapi-rb?

JSONAPI Suite supports deeply nested reads and writes, [integration spec helpers](https://jsonapi-suite.github.io/jsonapi_suite/how-to-write-specs), [automatic documentation](https://jsonapi-suite.github.io/jsonapi_suite/how-to-autodocument), and [clients](https://github.com/jsonapi-suite/jsorm) that provide a friendly ActiveRecord-like interface. I'd love to get a sense of your roadmap to see if we could use this as an alternate serialization layer.

grossadamm:
We have not, and it would certainly be worth doing!

minsikzzang:
benchmark against `jsonapi-rb` added https://github.com/Netflix/fast_jsonapi/pull/52

PR #52, opened 2018-02-05
Add jsonapi rb benchmark test

minsikzzang:
As mentioned in https://github.com/Netflix/fast_jsonapi/issues/12 and https://github.com/Netflix/fast_jsonapi/issues/44, added benchmark test with `jsonapi-rb`

It's hard to make direct comparison, because `jsonapi-rb` renders `relationships` only when we include.

e.g.) `AMS` / `fast_jsonapi` render movie class without including =>
```
  { 
    ...,
    "relationships" : {
      "actors" : {"data":[{"id":"1","type":"actor"} .... ]},
  }
```

`jsonapi-rb` render movie class without including => 
```
{
  ...,
  "relationships" : {
    "actors" : { "meta" : { "included" : false } .... ]},
}
```

* Benchmark results x 1000
```
Serialize to JSON string 1000 records
AMS serializer  1000       369.62 ms
jsonapi-rb serializer 1000       32.99 ms
Fast serializer 1000       10.31 ms


Serialize to JSON string 1000 with includes and meta
AMS serializer  1000       509.1 ms
jsonapi-rb serializer 1000       91.17 ms
Fast serializer 1000       20.82 ms
```


guilleiguaran:
Can you share benchmarks results here? 😊

NullVoxPopuli:
would it make sense to have additional comparisons?

with / without includes?

with / without sparse fieldsets?

with / without relationships (specifically belongs to / has many / both) ?

NullVoxPopuli:
I think it'd also be interesting to see how this interacts with ActiveRecord. Looking at the has_one 'extension', it looks like it's doing `object.relationship.id`, which may or may not cause a DB hit, which wouldn't be required if the foreign key is already on `object`, ya know/

bf4:
@grossadamm fwiw, I did some goofy shell comparison of serializations (jq and diff) in [https://github.com/rails-api/active_model_serializers/blob/benchmarks/benchmarks/serialization_libraries/compare.sh](https://github.com/rails-api/active_model_serializers/blob/b30b8cae7388ef1554e484a14c903b44ff41c1c5/benchmarks/serialization_libraries/compare.sh)

NullVoxPopuli:
The shell comparison are super valuable. When I initially did a jsonapi-rb vs AMS comparison, I didn't take into consideration the different default outputs.

A true comparison should have as-equal-as-possible json outputs.

shishirmk:
@minsikzzang Thank you this is super helpful.

PR #54, opened 2018-02-05
ability to customize rendering of attributes via a block

### LINKED ISSUE #5: Custom attributes

christophersansone:
Fixes #5.  Solves the same problem as #49 but with a minimal code change.  The serialization declaration is somewhat different as well.

Example:

```ruby
class MovieSerializer

  attribute :title_with_year do |record|
    "#{record.name} (#{record.release_year})"
  end
end
```


minsikzzang:
I prefer this solution than other one 👍  I might have to say the other solution could cause little bit of performance degradation.

christophersansone:
@grossadamm Good question.  Currently `attribute` and `attributes` are aliased to the same method, so they are completely interchangeable.  AMS does it this way, and I have seen that in other gems too.  It's really just syntactic sugar.  This works too:

```ruby
  attribute :name
  attributes :release_year do
    ...
  end

  attributes :actor_names, :starring do |record|
    record.actors.map(&:name).join(', ')
  end
```

Not sure why you would ever want to have multiple attributes evaluate to the same block, but it's possible... maybe if you're using `Object#send`...

apneadiving:
I like this implementation, makes complete sense since we try to avoid instantiating serializer objects. Maybe the attribute alias should be in another PR though

shishirmk:
@christophersansone please resolve the conflicts so that I can merge this. Could you also add to documentation an example of how to use the custom attributes feature with blocks?

Thank you for being patient and helping with trying out all the different ideas.

minsikzzang:
I think you have to rebase on `dev`

christophersansone:
@shishirmk Ready to go!

minsikzzang:
👏

PR #56, opened 2018-02-05
Better usage of AMS::Model

bf4:
Nothing particularly intereting, just, if we're using AMS::Model,
might as well use it as designed.

This also seems to break two of the benchmarks on my machine, making AMS
not quite only 22-24x slower.

I'm not sure whether the benchmark is intended more for regression testing than for a goal.
I could also add benchmarks for the other **two** major AMS releases (:facepalm:).

Note: I am an AMS maintainer, though it isn't much being developed these days.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -94,7 +94,7 @@ PLATFORMS
   ruby
 
 DEPENDENCIES
-  active_model_serializers (~> 0.10.4)
+  active_model_serializers (~> 0.10.7)
   activerecord (~> 5.0)
   bundler (~> 1.0)
   byebug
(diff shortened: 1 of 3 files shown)


shishirmk:
I changed the pull request to dev as there are many changes there that we havent merged into master yet. 

The performance tests are mostly to catch performance regressions. Here is an example of how they help with design decisions. https://github.com/Netflix/fast_jsonapi/pull/49

bf4:
Thanks. Rebased. I also added a commit to correct the rubygems.org url.  Wasn't sure if you'd want that as a separate PR.

shishirmk:
Thank you for rebasing. Merging it in.

bf4:
wow, quick merge!  Working late!

re:

> The performance tests are mostly to catch performance regressions. 

Well, in that case, AMS isn't under particularly active development, so I guess could be a stable benchmark. I've been interested in your benchmarking code, as I wrote some fancy stuff for AMS, but didn't end up using it much.  I like how it's integrated and doesn't take *too* long.

PR #58, opened 2018-02-06
RFC: Let people choose their JSON serializer; JSON:API is not circular

bf4:
For discussion:

- MultiJSON has historically had performance issues, and, in my opinion, is an unnecessary runtime dependency.
  It's just as easy to tell people to write their own `to_json` method if they'd
  like to use something other than the standard library.
- `JSON.fast_generate` is like `JSON.dump` but doesn't check for circle in Ruby objects.
  Since we're always serializing a flattened Hash, that shouldn't be an issue.
  It also isn't overriden by gems like ActiveSupport or oj_mimic_json like `to_json` is,
  so there's less uncertainty about what is actually doing the serialization.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,7 +3,6 @@ PATH
   specs:
     fast_jsonapi (1.0.17)
       activesupport (>= 4.2)
-      multi_json (~> 1.12)
 
 GEM
   remote: https://rubygems.org/
@@ -61,7 +60,6 @@ GEM
       nokogiri (>= 1.5.9)
     mini_portile2 (2.3.0)
     minitest (5.10.3)
-    multi_json (1.13.1)
     nokogiri (1.8.1)
       mini_portile2 (~> 2.3.0)
     oj (3.4.0)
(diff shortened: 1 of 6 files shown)


shishirmk:
We used multi_json and OJ in our internal services so it made sense to use them for this gem as well. The current dev branch removed the runtime dependency of OJ. #40 was the pull request which removed some other dependencies too.

Could you add documentation for how someone could use OJ if they wanted to. We were using multi_json so that this configuration can be handled by multi_json.

bf4:
I went a little overboard in the multi-json removal here... gotta think about this or if I should just make a new gem

shishirmk:
Wow looks like you are right this should be it's own gem. I will merge in the changes for now. I like that it reduces a dependency.

bf4:
Oh, thanks @shishirmk I would have written some tests for the different gems.  I mostly just tested by commenting things out :)

bf4:
Well, I went ahead and did it https://twitter.com/hazula/status/961475019363835904 https://github.com/bf4/fast_multi_json and while there also made https://github.com/bf4/safety_dance (ref  https://github.com/intridea/multi_json/pull/113 )

martndemus:
This should have been a major version bump. My whole test suite glowed up red when updating this gem to it's latest version, which turns out to be caused by this PR.

PR #61, opened 2018-02-06
Proof of concept for `ActiveSupport::Notifications`

hmcfletch:
This is working towards generic Skylight integration. Been wanting to take a crack at one of the skylight integrations, figured this was a good opportunity.

This is just a first pass. Opening the PR now to get feedback.

### DIFF
--- a/lib/fast_jsonapi/object_serializer.rb
+++ b/lib/fast_jsonapi/object_serializer.rb
@@ -18,7 +18,8 @@ module FastJsonapi
     end
 
     SERIALIZE_HASH_NOTIFICATION = 'render.fast_jsonapi.serializable_hash'.freeze
-    TO_JSON_HASH_NOTIFICATION = 'render.fast_jsonapi.to_json'.freeze
+    SERIALIZED_JSON_NOTIFICATION = 'render.fast_jsonapi.serialized_json'.freeze
+    TO_JSON_NOTIFICATION = 'render.fast_jsonapi.to_json'.freeze
 
     extend ActiveSupport::Concern
     include SerializationCore
@@ -32,7 +33,7 @@ module FastJsonapi
 
         def to_json
           if instrumentation_enabled?
-            ActiveSupport::Notifications.instrument(TO_JSON_HASH_NOTIFICATION, { name: self.class.name }) do
+            ActiveSupport::Notifications.instrument(TO_JSON_NOTIFICATION, { name: self.class.name }) do
               to_json_without_instrumentation
             end
           else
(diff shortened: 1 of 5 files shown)


hmcfletch:
Still needs to work in the Skylight stuff. 

We could also optionally define the method based on if the instrumentation is enabled or not. Would prevent a bunch of conditional checks.

lib/fast_jsonapi/object_serializer.rb
```diff
@@ -1,36 +1,45 @@
...
-        instrument_method :to_json
-      end
-
       # Set record_type based on the name of the serializer class
       set_type default_record_type if default_record_type
+
+      self.class_eval do
+        alias_method :to_json_without_instrumentation, :to_json
+
+        def to_json
```
shishirmk: we dont have a method called `to_json`. I think the method we use is called `serialized_json` to get the JSON string representation. 

I like the simple and readable approach you have taken with `serializable_hash` can we do the same with `serialized_json`?

spec/lib/object_serializer_notifications_spec.rb
```diff
@@ -0,0 +1,59 @@
+require 'spec_helper'
+
+describe FastJsonapi::ObjectSerializer do
+  include_context 'movie class'
+
+  context 'instrument' do
```
shishirmk: it would be more readable if the spec were named a little elaborately. It's important to test both cases.
```
context 'when calling serializable_hash' do
  it 'should send a notification' do
  end
  
  it  'shouldnt send a notification' do
  end
end
```

lib/fast_jsonapi/object_serializer.rb
hmcfletch: I was going off the original Skylight instrumenting, [here](https://github.com/Netflix/fast_jsonapi/pull/61/files#diff-34088120ad9556cc6578b0c51b38c723L29), which did `serializable_hash` and `to_json`. It would be easy enough to change.

spec/lib/object_serializer_notifications_spec.rb
hmcfletch: Of course. These tests were mostly just to make sure that I was actually getting notifications. :p 

TDD FTW!!!

wagenet:
For Skylight, you can make a normalizer. The normalizer is a class that extends `Skylight::Normalizers::Normalizer`. In the class you'll call the `register` method for each AS::N instrumentation you want to track. When the corresponding instrumentation is hit in AS::N, the `normalize` method is called. This takes 3 arguments: the current Skylight trace, the name of the AS::N instrumentation, and the AS::N payload. The expected response is an array of `[category, title, description]` which are all the same as what you would pass to `Skylight.instrument`. If, for some reason, you don't want to instrument that request, you can return `:skip` instead.

Hopefully, that all made sense. I really should write some proper documentation for this!

(Also, there will be some minor changes to class names for Skylight 2.0, as well as some new available hooks. However, existing code should basically work.)

wagenet:
See an example normalizer here: https://github.com/skylightio/skylight-ruby/blob/v1.5.0/lib/skylight/normalizers/active_record/sql.rb. You can also find others in the normalizers directory.

hmcfletch:
Would probably prefer #88 over this version.

shishirmk:
@hmcfletch should we close this?

hmcfletch:
yeah, this was the initial work for #88

Issue #57, opened 2018-02-06
`fields` option to choose what attributes will be used in serialization


youroff:
Would be awesome if it was possible to choose what keys will be used to generate json. So that depending on context, using the same serializer, we could render light-version of the model. This is what we used a lot in AMS. If this makes sense, I can provide a PR.

shishirmk:
@youroff I would be happy to look at PR. This is an enhancement that we would like to see in the gem.

nurey:
Would love to see this. This is JSON API sparse fieldsets. http://jsonapi.org/format/#fetching-sparse-fieldsets

youroff:
There's a couple of problems though. On my computer perf tests are randomly failing 50% of times. Which happens without modifications too though.

I'm not sure that it will work if different key_transform was used in different serializers. To fix that, we might instantiate serializers when processing included records, instead of using their internals on the class level. It might affect performance, but bring some safety.

shishirmk:
This feature is released in [version 1.3](https://github.com/Netflix/fast_jsonapi/releases/tag/1.3)

PR #63, opened 2018-02-06
Sparse fieldsets support, addresses #57

youroff:
This PR addresses #57 
It provides support for sparse fieldsets per json-api spec. Usage:
```ruby
options = {}
options[:meta] = { total: 2 }
options[:include] = [:actors]
options[:fields] = {
  movie: ['name'], # will only include name attribute into attributes hash
  actor: ['email'] # ...only email for included actors
}
MovieSerializer.new([movie, movie], options).serialized_json
```

### DIFF
--- a/lib/fast_jsonapi/serialization_core.rb
+++ b/lib/fast_jsonapi/serialization_core.rb
@@ -26,9 +26,17 @@ module FastJsonapi
         id_hash(ids, record_type) # ids variable is just a single id here
       end
 
-      def attributes_hash(record)
-        attributes_to_serialize.each_with_object({}) do |(key, method_name), attr_hash|
-          attr_hash[key] = record.public_send(method_name)
+      def attributes_hash(record, fields = nil)
+        if fields.is_a? Array
+          fields.each_with_object({}) do |key, attr_hash|
+            if method_name = attributes_to_serialize[key]
+              attr_hash[key] = record.public_send(method_name)
+            end
+          end
+        else
+          attributes_to_serialize.each_with_object({}) do |(key, method_name), attr_hash|
+            attr_hash[key] = record.public_send(method_name)
+          end
         end
       end
 
@@ -46,11 +54,11 @@ module FastJsonapi
         end
       end
 
-      def record_hash(record)
+      def record_hash(record, fields = nil)
         if cached
           record_hash = Rails.cache.fetch(record.cache_key, expires_in: cache_length) do
             temp_hash = id_hash(record.id, record_type) || { id: nil, type: record_type }
-            temp_hash[:attributes] = attributes_hash(record) if attributes_to_serialize.present?
+            temp_hash[:attributes] = attributes_hash(record, fields) if attributes_to_serialize.present?
             temp_hash[:relationships] = {}
             temp_hash[:relationships] = relationships_hash(record, cachable_relationships_to_serialize) if cachable_relationships_to_serialize.present?
             temp_hash
@@ -59,7 +67,7 @@ module FastJsonapi
           record_hash
         else
           record_hash = id_hash(record.id, record_type) || { id: nil, type: record_type }
-          record_hash[:attributes] = attributes_hash(record) if attributes_to_serialize.present?
+          record_hash[:attributes] = attributes_hash(record, fields) if attributes_to_serialize.present?
           record_hash[:relationships] = relationships_hash(record) if relationships_to_serialize.present?
           record_hash
         end
@@ -71,7 +79,7 @@ module FastJsonapi
 
       # includes handler
 
-      def get_included_records(record, includes_list, known_included_objects)
+      def get_included_records(record, includes_list, fields, known_included_objects)
         includes_list.each_with_object([]) do |item, included_records|
           object_method_name = @relationships_to_serialize[item][:object_method_name]
           record_type = @relationships_to_serialize[item][:record_type]
@@ -80,11 +88,12 @@ module FastJsonapi
           included_objects = record.send(object_method_name)
           next if included_objects.blank?
           included_objects = [included_objects] unless relationship_type == :has_many
+          included_fields = fields[record_type]
           included_objects.each do |inc_obj|
             code = "#{record_type}_#{inc_obj.id}"
             next if known_included_objects.key?(code)
             known_included_objects[code] = inc_obj
-            included_records << serializer.record_hash(inc_obj)
+            included_records << serializer.record_hash(inc_obj, included_fields)
           end
         end
       end
--- a/lib/fast_jsonapi/serialization_core.rb
+++ b/lib/fast_jsonapi/serialization_core.rb
@@ -40,17 +40,19 @@ module FastJsonapi
         end
       end
 
-      def relationships_hash(record, relationships = nil)
+      def relationships_hash(record, fields, relationships = nil)
         relationships = relationships_to_serialize if relationships.nil?
 
         relationships.each_with_object({}) do |(_k, relationship), hash|
           name = relationship[:key]
-          id_method_name = relationship[:id_method_name]
-          record_type = relationship[:record_type]
-          empty_case = relationship[:relationship_type] == :has_many ? [] : nil
-          hash[name] = {
-            data: ids_hash(record.public_send(id_method_name), record_type) || empty_case
-          }
+          if fields.nil? || fields.include?(name)
+            id_method_name = relationship[:id_method_name]
+            record_type = relationship[:record_type]
+            empty_case = relationship[:relationship_type] == :has_many ? [] : nil
+            hash[name] = {
+              data: ids_hash(record.public_send(id_method_name), record_type) || empty_case
+            }
+          end
         end
       end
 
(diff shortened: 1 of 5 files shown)


lib/fast_jsonapi/serialization_core.rb
```diff
@@ -26,9 +26,17 @@ def ids_hash(ids, record_type)
         id_hash(ids, record_type) # ids variable is just a single id here
       end
 
-      def attributes_hash(record)
-        attributes_to_serialize.each_with_object({}) do |(key, method_name), attr_hash|
-          attr_hash[key] = record.public_send(method_name)
+      def attributes_hash(record, fields = nil)
```
shishirmk: Sparse fieldsets are supposed to work with attributes and relationships as per the jsonapi spec. Here is the definition of fields http://jsonapi.org/format/#document-resource-object-fields

I think we would have to only return the relationships listed in the fieldsets. We would need to add changes for that

lib/fast_jsonapi/serialization_core.rb
```diff
@@ -26,9 +26,17 @@ def ids_hash(ids, record_type)
...
+        if fields.is_a? Array
```
shishirmk: This check is happening for every record. it would cause some performance degradation. Ideally this check should happen only once per serializer instantiation.

Do you mind posting the performance numbers you see when you run the benchmark after these changes?

spec/lib/object_serializer_spec.rb
```diff
@@ -36,6 +40,17 @@
       expect(serializable_hash['meta']).to be_instance_of(Hash)
     end
 
+    it 'returns correct fieldset when serializer instantiated with fields option' do
```
shishirmk: I think it makes sense to add a performance spec too. That would give us an idea of how this feature performs in fast_jsonapi compared to AMS.

shishirmk:
If i reading this correctly the model_name in the syntax could be plural or singular. Here is the link http://jsonapi.org/format/#fetching-sparse-fieldsets. I am not sure if this implementation takes care of that scenario.

So this would be allowed too?
```ruby
options = {}
options[:meta] = { total: 2 }
options[:include] = [:actors]
options[:fields] = {
  movie: ['name'], # will only include name attribute into attributes hash
  actors: ['email'] # ...only email for included actors
}
MovieSerializer.new([movie, movie], options).serialized_json
```

lib/fast_jsonapi/serialization_core.rb
youroff: If I understand what you mean, this might conflict with `include` option. Suppose you're returning main model A and models B and C in relationships. And you want sparse fieldsets in A and B, but C must be full. So if we were to treat keys from `fields` as explicit include, we would have to enumerate all fields there. But if we only treat fields as a guide of which fields to render per model, then we're fine if we don't have fieldset for some model, but it's included (this is how it works now).

lib/fast_jsonapi/serialization_core.rb
youroff: We can do something like this to avoid check:
```
        (fields || attributes_to_serialize.keys).each_with_object({}) do |key, attr_hash|
          if method_name = attributes_to_serialize[key]
            attr_hash[key] = record.public_send(method_name)
          end
        end
```
Or we can move fields-filtering logic to instantiation phase, but it probably won't work with relationships, since they're processed without instantiation of serializer.

spec/lib/object_serializer_spec.rb
youroff: Ok, will do that

youroff:
Here's what spec says: http://jsonapi.org/format/#document-resource-object-identification 
"Note: This spec is agnostic about inflection rules, so the value of type can be either plural or singular. However, the same value should be used consistently throughout an implementation."

And current implementation would expect record type as a key. actors would not be allowed, unless all record keys are plural. I suspect you're considering `fields` keys as path, but in fact it's a record types. In that example in jsonapi spec, they're including relationship `author`, but in fields, there's `people` key that defines how author will be serialized.

lib/fast_jsonapi/serialization_core.rb
shishirmk: @youroff I am ok with having the implicit assumption that if a model is not mentioned in the fieldsets then we its included.

I was actually asking about relationships. Let's take the example you have mentioned above. I will simplify it a bit.
```ruby
options = {}
options[:fields] = {
  movie: ['name'], # will only include name attribute into attributes hash
}
MovieSerializer.new([movie, movie], options).serialized_json
```
Let's assume this is our movie serializer
```ruby
class MovieSerializer
    belongs_to :movie_type
end
```

the output of the serializer would it have the movie_type relationship in the json? or would it only have the name attribute. If I understand the specification correctly it shouldn't have any relationships just the name attribute. However i dont think this implementation is taking care of that scenario.

youroff: Ah... Gotcha. Yeah, would make sense if relationships were `fields`-aware too. I'll fix that.

lib/fast_jsonapi/serialization_core.rb
shishirmk: your point about moving logic to instantiation phase makes sense. I am wondering if i should make record_hash, attributes_hash and relationships_hash instance methods.

youroff: Would it make sense to generalize approach so that it was possible to include relationships at any depth? For example, with `include` and `fields`:
```
options = {}
options[:include] = [:actors, movie_type: [:genre, ...]]
options[:fields] = {
  movie: ['name', 'actors'],
  actor: ['name'],
  movie_type: ['genre', 'something_else']
}
MovieSerializer.new([movie, movie], options).serialized_json
```
And during initialization phase we would instantiate all serializers in the tree, and instead of calling class level methods to get hashes for included records, we would call respective serializer instances. Not sure about performance implications, caching and duplicate prevention so far.

shishirmk: Yup i think it would makes sense to do that. I am worried about the performance impact too. In order to understand what the performance impact might be I will do a quick test.
I will make the three methods i mentioned instance methods and compare it to current performance.

Current Performance looks like this

```
Serialize to Ruby Hash 250 records
AMS serializer  250        104.49 ms
jsonapi-rb serializer 250        8.14 ms
Fast serializer 250        2.32 ms

Serialize to Ruby Hash 1000 records
AMS serializer  1000       348.31 ms
jsonapi-rb serializer 1000       36.72 ms
Fast serializer 1000       8.72 ms
```

youroff: Yeah, I was wondering about that too... Does really performance depends on the use of static methods that much?

shishirmk: Here is the performance after those three methods were converted to become instance methods

```
Serialize to Ruby Hash 250 records
AMS serializer  250        114.01 ms
jsonapi-rb serializer 250        7.89 ms
Fast serializer 250        2.23 ms

Serialize to Ruby Hash 1000 records
AMS serializer  1000       339.41 ms
jsonapi-rb serializer 1000       35.53 ms
Fast serializer 1000       9.53 ms
```

youroff: Can it be just some kind of fluctuation? For 250 records it turned out to be faster. How many times did you repeat the test?

youroff: This is weird... So I created this silly test that would in my understanding show the speed of method dispatch: https://gist.github.com/youroff/dc4f0ff71083acf6c180ec33f3c52881
And calling class methods is consistently slower. Maybe that's something else or I'm just missing something?

shishirmk: I repeated the test around 5 times. It did fluctuate quite a bit. Actually I think i am ok with the small degradation if it means easy to understand code and spare fieldsets and includes would be supported better.

shishirmk: I saw that calling class method was slower too. but i think if its a instance method i think the instance size goes up and that causes some changes in speed too i guess.

youroff: Another guess... If you just moved methods to the instance, but they kept using class level attributes, maybe their dispatch slowed down the execution? Anyways, I have an idea how this can be arranged, but I need to experiment a bit. Will try to do that on the weekend.

youroff: @shishirmk so, I attempted to implement the strategy I was thinking about, but it turned out to be pretty much a complete rewrite. And so far, being just a proof of concept, it doesn't support some features that currently present in fast_jsonapi, but performance wise it seems to be the same give or take. I'm not sure if change like that can be a PR. Is there a way we could chat to figure out what to do with it?

shishirmk: Do you want to try using https://gitter.im/ to chat? I havent used it before but seems like a good place to chat https://gitter.im/fast_jsonapi/Lobby

ctm:
Any update on this?Is it blocked awaiting a performance spec of the feature for comparison to AMS, or are there other issues? I have a project used in production where we use AMS with sparse fieldsets and we're going to be ramping up use. I'm concerned that AMS is too slow (and also doesn't seem to be maintained).

ababich:
@shishirmk is there any plans to work on these or we should consider another implementation for the `fields` feature?

youroff:
@ctm @ababich @sethgw It turned out to be easier to rewrite the library, while keeping the initial idea and focus on performance. You may want to take a look at it:
https://github.com/youroff/jsonapi_serializer/

It implements fields and much more, such as arbitrary nesting, polymorphic relationships etc, while maintaining similar performance. Integration with rails is pretty much manual so far, though it may change if the library gets some interest.

shishirmk:
@ababich  @sethgw @ctm i would like look at this feature and merge it too. I don't have a date right now. but sparse fields and optional attributes are the features we are targeting for the next release.

sethgw:
@shishirmk @ababich @ctm 

this PR https://github.com/Netflix/fast_jsonapi/pull/254 is doing what I need!

shishirmk:
@sethgw left a comment there

sethgw:
I understand the problem better now after a few hours digging through things.  I will see if I can come up with a solution to have "fields" respect both attributes and relations.  I tried a few things, but it's going to take more work to get it right.

shishirmk:
#271 added this feature so closing this pull requst

PR #67, opened 2018-02-08
Feature/custom attributes

shishirmk:
This is an attempt to provide the custom attributes feature by using only a single serializer instance and using it as a decorator as well. 

@guilleiguaran, @christophersansone take a look at this and let me know what you think. The feature is complete but i have to still fix some tests and measure performance.

### DIFF
--- a/lib/fast_jsonapi/serialization_core.rb
+++ b/lib/fast_jsonapi/serialization_core.rb
@@ -15,6 +15,49 @@ module FastJsonapi
                       :cache_length,
                       :cached
       end
+
+      def klass
+        self.class
+      end
+
+      def attributes_hash(record)
+        klass.attributes_to_serialize.each_with_object({}) do |(key, method_name), attr_hash|
+          attr_hash[key] = record.public_send(method_name)
+        end
+      end
+
+      def relationships_hash(record, relationships = nil)
+        relationships = klass.relationships_to_serialize if relationships.nil?
+
+        relationships.each_with_object({}) do |(_k, relationship), hash|
+          name = relationship[:key]
+          id_method_name = relationship[:id_method_name]
+          record_type = relationship[:record_type]
+          empty_case = relationship[:relationship_type] == :has_many ? [] : nil
+          hash[name] = {
+            data: klass.ids_hash(record.public_send(id_method_name), record_type) || empty_case
+          }
+        end
+      end
+
+      def record_hash(record)
+        if klass.cached
+          record_hash = Rails.cache.fetch(record.cache_key, expires_in: klass.cache_length) do
+            temp_hash = klass.id_hash(record.id, klass.record_type) || { id: nil, type: klass.record_type }
+            temp_hash[:attributes] = attributes_hash(record) if klass.attributes_to_serialize.present?
+            temp_hash[:relationships] = {}
+            temp_hash[:relationships] = relationships_hash(record, klass.cachable_relationships_to_serialize) if klass.cachable_relationships_to_serialize.present?
+            temp_hash
+          end
+          record_hash[:relationships] = record_hash[:relationships].merge(relationships_hash(record, klass.uncachable_relationships_to_serialize)) if klass.uncachable_relationships_to_serialize.present?
+          record_hash
+        else
+          record_hash =klass.id_hash(record.id, klass.record_type) || { id: nil, type: klass.record_type }
+          record_hash[:attributes] = attributes_hash(record) if klass.attributes_to_serialize.present?
+          record_hash[:relationships] = relationships_hash(record) if klass.relationships_to_serialize.present?
+          record_hash
+        end
+      end
     end
 
     class_methods do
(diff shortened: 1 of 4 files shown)


lib/fast_jsonapi/serialization_core.rb
```diff
@@ -15,56 +15,63 @@ class << self
...
           }
         end
       end
 
       def record_hash(record)
-        if cached
-          record_hash = Rails.cache.fetch(record.cache_key, expires_in: cache_length) do
-            temp_hash = id_hash(record.id, record_type) || { id: nil, type: record_type }
-            temp_hash[:attributes] = attributes_hash(record) if attributes_to_serialize.present?
+        self.object = record
```
christophersansone: Consider making sure that `object` is only available within the context of this method, i.e.

```ruby
self.object = record
begin
  ...
ensure
  self.object = nil
end
```

christophersansone:
@shishirmk Thanks, looks good too.  This implementation avoids the need to instantiate a serializer for each record, and protected / private methods will work.  The challenge / concern would be that state (i.e. instance variables) would persist across all iterations, as opposed to starting clean each time.  Consider this:

```ruby
class FooSerializer
  attributes :a, :b

  def complex_calculation
    @complex_calculation ||= ...
  end

  def a
    complex_calculation.foo
  end

  def b
    complex_calculation.bar
  end
end
```

In this implementation, `complex_calculation` would be calculated only upon serialization of the first object.  Subsequent serializations would use the same value, even if the value was generated from `object`.  Is that a feature or a bug?  I guess it depends, but it is different than AMS.

Lastly, the `object` variable seems a bit "magical" in this implementation.  It exists outside the context of the method itself and is available to the method during the serialization process, but only during the serialization process.  If one were to call `MovieSerializer.new(movie).title_with_year`, it would fail.  (We could improve the default attribute definition method to throw a descriptive error if this occurs, which will help document this limitation.)

So, there are a couple side effects with this implementation.  Maybe people won't notice or care in the end, but those who wind up noticing these will see some unexpected behavior.  @guilleiguaran 's implementation has a similar side effect with private / protected methods, and his implementation will be slower and consume more memory due to the amount of instances being created.

That said, if we _reeeeeally_ feel the need to offer verbatim AMS syntax, these are the two options, each with their pros and cons.  If I had to choose between the two, I would choose the more performant solution (this one).  Personally, I would prioritize performance and a syntax that is more in-line with the architecture, over the need to support AMS syntax... but I realize I'm probably the minority.  Porting from AMS to FastJsonapi won't be verbatim no matter what we do here... and I think either way we should offer a guide describing specifically what needs to change...

christophersansone:
This got me thinking about (yet another) implementation idea.  What if we re-defined what an "instance" actually is?  Right now we call the following to start serialization:

```ruby
MovieSerializer.new(movies).serializable_hash
```

What if we changed that so that this occurs at the class level, e.g.

```ruby
MovieSerializer.serializable_hash(movies)
MovieSerializer.to_json(movies)
MovieSerializer.dump(movies)
```

Then, a single reusable instance will be created internally in this process.  From the developer's perspective, the job of an instance is to serialize a single object.  The `object` variable works like it does in AMS, private and protected methods work, and instance variables would be single use too.  How do we achieve single-use instance variables?  Through some sort of reset method that does something like this:

`
instance_variables.each { |v| instance_variable_set("@#{v}", nil) }
`

A pseudo-code implementation would look something like this:

```ruby
def self.serializable_hash(objects)
  if objects.is_a?(Array)
    serialize_multiple(objects)
  else
    serialize_single(objects)
  end
end

def self.serialize_multiple(objects)
  ...
  serializer_instance = self.new
  objects.each do |object|
    serializer_instance.serializable_hash(object)
    serializer_instance.reset_instance_variables
  end
end
```

I think this implementation would better draw the line between which methods belong at the class level and which methods belong at the instance level.

@shishirmk @guilleiguaran what do you think of this approach?  If you think it's worthwhile, I can try to implement later this evening.

guilleiguaran:
@christophersansone please do it!!!

I played with the idea of using a single serializer instance as decorator in https://github.com/guilleiguaran/fast_jsonapi/commit/1e717748ec67ea64ddde33eb134c0f6081222012 but I'm not really happy with the performance results of it:

dev branch:
```
Serialize to JSON string 10000 records
AMS serializer  10000      3204.8 ms
Fast serializer 10000      109.93 ms

Serialize to Ruby Hash 10000 records
AMS serializer  10000      3025.55 ms
Fast serializer 10000      90.17 ms
```

custom-attributes-v3 branch:
```
Serialize to JSON string 10000 records
AMS serializer  10000      3194.23 ms
Fast serializer 10000      133.75 ms

Serialize to Ruby Hash 10000 records
AMS serializer  10000      2982.17 ms
Fast serializer 10000      117.76 ms
```

shishirmk:
@guilleiguaran @christophersansone You guys have been awesome. Thank you for trying out all the different ideas we came up with. I think custom attributes is a very important feature a lot of people have asked for it. I would like to have it in the upcoming 1.1.0 release which we are planning to release next week. Keeping in mind the time we have, performance and ease of use I am planning to merge in #54. We can continue to refine the feature after we release 1.1.0

christophersansone:
@shishirmk @guilleiguaran Thanks team.  Good call on just merging in attribute blocks for now, since it gets the job done without all of these concerns.

I created #71 to extract this "class vs. instance" discussion.  Once we have some decisions made there and have an implementation available, custom attributes and many other features will fall into place.

PR #69, opened 2018-02-09
Update README.md

ritaly:
That's quite important info ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Build Status](https://travis-ci.org/Netflix/fast_jsonapi.svg?branch=master)](https://travis-ci.org/Netflix/fast_jsonapi)
 
-A lightning fast [JSON:API](http://jsonapi.org/) serializer for Ruby Objects.
+A lightning fast [JSON:API](http://jsonapi.org/) serializer for Ruby Objects. Rails version 5.0+ is required.
 
 # Performance Comparison
 


guilleiguaran:
Rails 4.2 is supported also: https://github.com/Netflix/fast_jsonapi/blob/dev/fast_jsonapi.gemspec#L23

ritaly:
yes, but we've got an error with active record version :)

guilleiguaran:
@ritaly can you try this in your Gemfile:

```ruby
gem 'fast_jsonapi', github: 'Netflix/fast_jsonapi', branch: 'dev'
```

and report here if it works correctly.

mknycha:
Thanks for you answer. I tried to install the gem from the default source, but got the error:
```
Bundler could not find compatible versions for gem "activerecord":
  In snapshot (Gemfile.lock):
    activerecord (= 4.2.10)

  In Gemfile:
    fast_jsonapi was resolved to 1.0.17, which depends on
      activerecord (~> 5.0)

    kaminari-bootstrap was resolved to 3.0.1, which depends on
      rails was resolved to 4.2.10, which depends on
        activerecord (= 4.2.10)

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.
```
It turned out that after `gem 'fast_jsonapi', github: 'Netflix/fast_jsonapi', branch: 'dev'` gem was installed correctly. In that case, could you please update gem version in [ruby gem source](https://rubygems.org/gems/fast_jsonapi/versions/1.0.17), or write in README which source to use?

mknycha:
Or are you planning to merge 'dev' branch into master, so that Rails 4.2 will be supported in the next release? We just want to make sure we are using the supported version of this gem.

guilleiguaran:
@mknycha that's right, Rails 4.2 will be supported in next release.

mknycha:
Ok thank you.

shishirmk:
Thanks for responding @guilleiguaran @mknycha the plan is to create a release branch next week and have a release out and them merge to master following the git flow https://jeffkreeftmeijer.com/git-flow/ for our releases.

PR #70, opened 2018-02-09
Documentation for polymorphic association

vovimayhem:
# What does this PR do?

* Adds documentation about how to set up polymorphic associations


vovimayhem:
I didn't knew how to document an option that can take either a `true` or a `Hash` with object class -> `record_type`, so I added two rows to the table.

vovimayhem:
@shishirmk On hindsight, I'm thinking that the `polymorphic` customization in the serializer DSL could be better... I'll start another discussion for this.

vovimayhem:
...see #72

shishirmk:
This should have probably been part of `1.1.0`. I will merge it in soon

PR #74, opened 2018-02-10
More intuitive DSL for homogeneous/heterogeneous associations

vovimayhem:
## What does this PR do?

This PR changes the way the serialization DSL supports heterogeneous associations (a.k.a. polymorphic). This comes with some breaking changes with the performance obtained by default.

* The serialization code will assume all associations to be heterogeneous (of different types), unless you explicitly configure this using the `record_type: :some_type` option.

* Also, the serialization code won't assume there is a `association_id` or `association_ids` method, and will check if each record class implements it or not.

Here is the new usage:

* By default, the serializer will support heterogeneous associations:

```ruby
class ProjectSerializer
  has_one :assignee  # Assignee can be of any type, including :user or :administrator
  has_many :tasks    # Tasks can be of any type, including :small_task, :task, or :big_tasks
end
```

* To "opt-in" for the performance benefits of having an "Homogeneous Association", you'll need to specify the `record_type` used by the association in the serializer:

```ruby
class ProjectSerializer
  has_one :assignee, record_type: :user       # Assignee will be rendered with type :user exclusively
  has_many :tasks,   record_type: :small_task # Tasks will be rendered with type :small_task exclusively
end
```

* If your associations are "homogeneous" and the object to serialize implements the `assignee_id` or `task_ids`, the serializer code will use those instead to generate the resource identifiers.

Under the hood, the `record_type` is now a `Hash` object by default, and it allows adding your own record type mappings for each associated object class, in case :

```ruby
class ProjectSerializer
  has_many :tasks, record_type: { SmallTask => :small_task, BigTask => :big_task }
end
```


vovimayhem:
Oops, some tests failing...

lib/fast_jsonapi/object_serializer.rb
```diff
@@ -170,68 +176,61 @@ def has_many(relationship_name, options = {})
...
-        return false unless option.present?
+      def fetch_record_type_option(options)
+        option = options[:record_type]
+        # Return it right away if it's a symbol or a Hash:
+        return option if option.is_a?(Symbol) || option.respond_to?(:keys)
+        # Return an empty Hash so the autodetect + memoization kicks in
+        {}
+      end
+
+      def fetch_record_type_option(options)
```
shishirmk: This method an the one above seem to have the same method signature. May be you meant to delete one?

vovimayhem: Indeed... last-minute copy & paste error... fixed!

lib/fast_jsonapi/object_serializer.rb
```diff
@@ -170,68 +176,53 @@ def has_many(relationship_name, options = {})
...
           serializer: compute_serializer_name(serializer_key),
-          relationship_type: :has_one,
-          cached: options[:cached] || false,
-          polymorphic: fetch_polymorphic_option(options)
+          is_collection: false,
+          cached: options[:cached] || false
         })
       end
 
+      alias belongs_to has_one
```
guilleiguaran: 👍

shishirmk:
We have diverged too much so closing this

PR #88, opened 2018-02-17
Instrumentation - Skylight Integration

hmcfletch:
I took another stab at the Skylight integration. I decided two have an explicit `require` to turn on the instrumentation instead of using environment variables or messing too much with the core files. This seemed a lot cleaner.


hmcfletch:
A few other notes:
1) Having some issues with the specs since if the spec for instrumentation gets run before the negative specs get run, then we get failing tests.
2) Not sure how to test the normalizers (and haven't tested them yet)
3) Looks like some of the stuff from master got in to this branch...

@wagenet Would love for you to take a look at what I have and see if anything you'd suggest.

shishirmk:
Thanks for looking into it @wagenet. @hmcfletch would love to try this out in our internal systems once it is merged to dev.

hmcfletch:
@shishirmk I updated the travis to run the tests separately. Everything is passing now.

.travis.yml
```diff
@@ -6,3 +6,4 @@ rvm:
   - 2.5.0
 script:
   - bundle exec rspec
+  - bundle exec rspec spec/lib/instrumentation/as_notifications.rb
```
GeekOnCoffee: If breaking out the spec is the only way to do it, chaining them with `&&` might be preferable (https://docs.travis-ci.com/user/customizing-the-build/#Customizing-the-Build-Step) as it would always generate testing output from both sets but still fail properly. 

That said, I think your test tear-down could alias the method back to the original and remove the newly added method easily enough and eliminate the need for the second running of rspec.

hmcfletch:
@GeekOnCoffee Took a crack at it. Hadn't done the `remove_method` thing before. Pretty nifty.

Issue #91, opened 2018-02-20
Allow includes to be strings


martndemus:
Having a `Series` serializer with `has_many :episodes`

```ruby
SeriesSerializer.new(series, include: ['episodes'])
```

Throws an `ArgumentError`.

I would expect it to work, since includes are often pulled from a controller's params, which are always strings.

Having `.to_sym` the strings would also be a memory issue in Ruby before 2.2.

sjmog:
Local pull request created [here](https://github.com/sjmog/fast_jsonapi/pull/1) that seems to demonstrate this error.

shishirmk:
Released as a part of version 1.1.0

PR #92, opened 2018-02-20
Make an error that demonstrates https://github.com/Netflix/fast_jsonapi/issues/91

sjmog:
* #91 suggests includes cannot be strings
* Here's a simple RSpec test that fails with a non-empty string but passes with a
non-empty symbol
* To run the test, `rspec spec/lib/object_serializer_spec.rb`


sjmog:
Sorry, I meant to keep this local to my fork (demonstration PR)

PR #93, opened 2018-02-20
91 allow includes strings

### LINKED ISSUE #91: Allow includes to be strings

sjmog:
Closes #91 

* Includes would fail with an ArgumentError unless they were explicitly provided as symbols (see #91).
* This is solved by mapping the strings to symbols in the `ObjectSerializer` initializer.
* No real impact on performance.


sjmog:
@guilleiguaran done

PR #95, opened 2018-02-21
Fix gemspec, README.rdoc isn't present in the project

guilleiguaran:
Trying to build the gem in `dev` branch I got an error:

```
~/src/fast_jsonapi[dev] % gem build fast_jsonapi.gemspec
WARNING:  See http://guides.rubygems.org/specification-reference/ for help
ERROR:  While executing gem ... (Gem::InvalidSpecificationException)
    ["README.rdoc"] are not files
```

The reason is that the project doesn't have a README.rdoc but the file is listed in the gemspec so I've removed it from gemspec.


shishirmk:
Thanks @guilleiguaran

PR #103, opened 2018-02-26
Title fixed for one spec

vlado:
Title for spec testing that namespaced class gets proper default record_type was wrong (copied from previous spec).

https://github.com/Netflix/fast_jsonapi/blob/master/spec/lib/object_serializer_spec.rb#L117

### DIFF
--- a/OSSMETADATA
+++ b/OSSMETADATA
@@ -1,0 +1,1 @@
+osslifecycle=active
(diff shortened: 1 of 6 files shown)


GeekOnCoffee:
@vlado looks like you've got some extra commits in here

vlado:
Thanks @GeekOnCoffee 

I've opened new PR with singe commit :)
https://github.com/Netflix/fast_jsonapi/pull/114

PR #110, opened 2018-03-10
Adding some specs for #as_json on ObjectSerializer class.

corinnekunze:
Added some tests here to ensure that we are maintaining availability of the `as_json` method on our Object Serializer.

Allowing `as_json` to successfully parse a json hash on the serializer was added in this pull request:
https://github.com/Netflix/fast_jsonapi/pull/89

I added a director attr_accessor in the movie context that acts to prove if an attribute is present on the record but not defined in the serializer, it should not be returned.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,9 @@ doc
 
 # For the gem
 test.db
+
+# For those using rbenv
+.ruby-version
+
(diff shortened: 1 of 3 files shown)


shishirmk:
Thank you for adding the tests

corinnekunze:
You bet!

PR #114, opened 2018-03-11
Title fixed for one spec

vlado:
Title for spec testing that namespaced class gets proper default record_type was wrong (copied from previous spec).

https://github.com/Netflix/fast_jsonapi/blob/master/spec/lib/object_serializer_spec.rb#L117


shishirmk:
Thanks for the fix

Issue #113, opened 2018-03-11
How to override the id attribute ?


Ranhiru:
According to the JSON API specifications there must be an `id` field. But I do not want to expose the `id` column of my DB and I have a `friendly_id` column (which is unique as well).

How do I instruct fast_jsonapi to use my `friendly_id` column instead ?

axsuul:
https://github.com/Netflix/fast_jsonapi/pull/54 enables you to render custom attributes and was recently merged in. 

You can point your `Gemfile` to the `dev` branch of this repo to use it.

shishirmk:
This is not supported today. It should be relatively small change to the code are you willing to make a pull request?

shuheiktgw:
PRed! https://github.com/Netflix/fast_jsonapi/pull/122

shishirmk:
Released as a part of version 1.1.0

PR #122, opened 2018-03-14
Enable to set id

shuheiktgw:
This PR relates to https://github.com/Netflix/fast_jsonapi/issues/113.
I implemented `FastJsonapi::ObjectSerializer#set_id` in order to set arbitrary column as `id`.

Thank you! 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ end
 class MovieSerializer
   include FastJsonapi::ObjectSerializer
   set_type :movie  # optional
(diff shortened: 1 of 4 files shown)


shishirmk:
Could you update the documentation to show an example of how to use set_id? I would be nice if you include an example of using set_id inside the serializer class definition

shuheiktgw:
I just added document! Thank you for pointing it out!

PR #125, opened 2018-03-16
fix behaviour for struct without id

moofkit:
```ruby
# test_id.rb
require "fast_jsonapi"

StructWithoutId = Struct.new(:name, :text)

class StructWithoutIdSerializer

  attributes :name, :text
end

StructWithoutIdSerializer.new(StructWithoutId.new("test", "test")).serializable_hash
```

```bash
ruby test_id.rb
=> undefined method `id' for #<struct StructWithoutId name="test", text="test"> (NoMethodError)
```
This pull requests adds exception with explanation why id is mandatory field.

### DIFF
--- a/lib/fast_jsonapi/serialization_core.rb
+++ b/lib/fast_jsonapi/serialization_core.rb
@@ -4,6 +4,8 @@ require 'active_support/concern'
 require 'fast_jsonapi/multi_to_json'
 
 module FastJsonapi
+  MandatoryField = Class.new(StandardError)
+
   module SerializationCore
     extend ActiveSupport::Concern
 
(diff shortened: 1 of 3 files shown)


shishirmk:
id is a mandatory field in the jsonapi spec look at jsonapi.org

moofkit:
@shishirmk I see...
WDYT about to add exception with this comment that you provide here? And add this to documentation why id is mandatory field.

shishirmk:
Documentation may not be necessary as that is part of a jsonapi spec. However giving a better exception message might be a good idea.

moofkit:
I add exception for this situation

shishirmk:
@moofkit  thank you

Issue #129, opened 2018-03-19
LoadError (cannot load such file -- yajl)


GarrisonD:
Hello, guys!

I upgraded `fast_jsonapi` gem to the latest version that was released recently (v1.1) and didn't update anything else. As a result I have such error:

`LoadError (cannot load such file -- yajl)`

GarrisonD:
As far as I understand, the error is here: https://github.com/Netflix/fast_jsonapi/blob/cf86cd871ecfe411d9c0fb22c85212b516a22138/lib/fast_jsonapi/multi_to_json.rb#L52

zabodaevmm:
+
switching to 1.0.17 not to reproduce this error

GarrisonD:
@zabodaevmm yeah, it's a workaround - but i like only solutions 😉

GarrisonD:
If someone else run into this trouble and #130 is not still merged, then you can manually add `oj` gem to your project dependencies and enjoy features from `fast_jsonapi-v1.1`!

shishirmk:
Looks like this needs to be a hot fix. @GarrisonD thank you for creating a pull request. I will look into and merge it as soon as i get a chance.

shishirmk:
just released 1.1.1 with the above changes

GarrisonD:
@shishirmk it's great news! thank you a lot!

bf4:
@GarrisonD sorry about that. I wrote that bug :)

PR #130, opened 2018-03-19
Fix bug with chaining rescues of FastJsonapi::MultiToJson::Result

GarrisonD:
Fix for #129.

### DIFF
--- a/lib/fast_jsonapi/multi_to_json.rb
+++ b/lib/fast_jsonapi/multi_to_json.rb
@@ -12,7 +12,12 @@ module FastJsonapi
     # e.g. https://github.com/github/github-ds/blob/fbda5389711edfb4c10b6c6bad19311dfcb1bac1/lib/github/result.rb
     class Result
       def initialize(*rescued_exceptions)
-        rescued_exceptions = [StandardError] if rescued_exceptions.empty?
+        @rescued_exceptions = if rescued_exceptions.empty?
+          [StandardError]
+        else
+          rescued_exceptions
+        end
+
         @value = yield
         @error = nil
       rescue *rescued_exceptions => e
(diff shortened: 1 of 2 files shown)


lib/fast_jsonapi/multi_to_json.rb
```diff
@@ -33,7 +38,8 @@ def value!
 
       def rescue
         return self if ok?
-        Result.new { yield(@error) }
+
+        Result.new(*@rescued_exceptions) { yield(@error) }
```
GarrisonD: Actually the bug was here - after the first `rescue` only `StandardError` was catched.

GarrisonD:
@shishirmk I hope you notice this PR 🙂 

And I think it would be nice to tell in `README.md` that users should install `oj` gem or any other JSON serializer as well to enjoy the power of `fast_jsonapi`.

shishirmk:
@GarrisonD it does make sense to add that to the README.

lib/fast_jsonapi/multi_to_json.rb
bf4: oops :)

```ruby
LoadError.superclass
=> ScriptError
>> LoadError.superclass.superclass
=> Exception

>> StandardError.superclass
=> Exception
```

GarrisonD: @bf4 oops what? 🙂 Looks like you don't know how `begin-rescue` construction works.

`rescue` catches instances of `StandardError` and its __inheritors__ but NOT __predecessors__.

So, I can't understand why you are talking about `.superclass` 🙁

bf4: @GarrisonD no, I know how it works. That's why I wrote oops :)

As you pointed out, rescuing StandardError doesn't rescue a LoadError. (That's the relevance of superclass. If LoadError.ancestors.include?(StandardError) there would have been no oops, right?)

Perhaps what wasn't clear was that I wrote the code with the bug and was commenting on *my* oops

GarrisonD: @bf4 Now everything is clear! Sorry.

PR #134, opened 2018-03-22
Refactor object serializer class methods spec

shuheiktgw:
I refactored `object_serializer_class_methods_spec.rb` with `subject`, `let` and `shared_example` since some of the examples seem to be not well organized... Also, I added some missing test cases and fixed seemingly wrong examples!

### DIFF
--- a/spec/lib/object_serializer_class_methods_spec.rb
+++ b/spec/lib/object_serializer_class_methods_spec.rb
@@ -4,73 +4,94 @@ describe FastJsonapi::ObjectSerializer do
 
   include_context 'movie class'
 
-  context 'when testing class methods of object serializer' do
+  describe '#has_many' do
+    subject(:relationship) { serializer.relationships_to_serialize[:roles] }
 
-    before(:example) do
-      MovieSerializer.relationships_to_serialize = {}
+    before do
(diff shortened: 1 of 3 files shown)


spec/lib/object_serializer_class_methods_spec.rb
```diff
@@ -4,72 +4,94 @@
...
-      expect(relationship[:record_type]).to eq 'area'.singularize.to_sym
+  describe '#belongs_to' do
+    subject(:relationship) { MovieSerializer.relationships_to_serialize[:area] }
+
+    before do
+      MovieSerializer.belongs_to *parent
     end
 
-    it 'returns correct relationship hash for a belongs_to relationship with overrides' do
-      MovieSerializer.has_many :area, id_method_name: :blah_id, record_type: :awesome_area, serializer: :my_area
```
shuheiktgw: I guess `has_many` was supposed to be `belongs_to` from the `it`  statement...

shuheiktgw:
I 'm also thinking about integrating `object_serializer_set_id_spec.rb` and `object_serializer_with_attribute_block_spec.rb` to this spec file after this PR is marged.

PR #144, opened 2018-03-29
Integrate class methods tests

shuheiktgw:
I integrated the test for `#set_id` and `#attribute` with block to object_serializer_class_methods_spec.rb. Also, I found I need to add some cleanups after executing tests so added them! Thank you! 👍


lib/fast_jsonapi/object_serializer.rb
```diff
@@ -111,11 +111,11 @@ def set_key_transform(transform_name)
           dash: :dasherize,
           underscore: :underscore
         }
-        @transform_method = mapping[transform_name.to_sym]
+        self.transform_method = mapping[transform_name.to_sym]
```
shuheiktgw: I found it is odd that every other class instance variable is accessed via accessor and only this one is not. So I changed it to accessor!

Issue #145, opened 2018-03-29
Add scope current_user like devise to check condition 


Phathdt:
Would you mind if have option/scope for current_user to check condition and render ? 
example we have model User with attribute name, age, dob, secret_key
then we check current user 
if current user is normal => just render name, age, dob
if current user is admin => render name, age, dob, secret_key
and condition we can use some gem like rolify, pundit
any one know that?

shishirmk:
fast_jsonapi doesn't have this functionality today of scopes. We would need to build it. Similar issue discussed in #139

Issue #139, opened 2018-03-23
Add #serialization_scope to enable context aware serialization


jdongelmans:
Would it be nice to have the ability 

https://github.com/rails-api/active_model_serializers/blob/0-10-stable/docs/general/serializers.md#scope

shishirmk:
@jdongelmans as a workaround you could put this logic in the model and it would work as expected

PR #150, opened 2018-03-31
Add scope to be able to serialize context aware

fongfan999:
This PR relates to #139 and #145


shishirmk:
How do you generally set scope? I am wondering if this should be something that is set in the serializer itself just like key_transform etc? 

Does the way you set scope change every time you invoke the serializer or is there a convention that is followed to derive the scope?

fongfan999:
Hi @shishirmk, 

The `scope` is generally `current_user`, `view_context`, ... And of course, it will not change every time we invoke the serializer.

We can pass `scope` as options on the fly like this:

```ruby
class MovieSerializer
  
  attribute :name do |object|
    # We can access the `:scope` as here
    scope.admin? ? object.admin_name : object.name
  end
end

options = {}
options[:meta] = { total: 2 }
options[:include] = current_user

```

Please let me know if it didn't make sense. Thank you

mhluska:
Are there any updates on this one? It looks like the merge conflict is an easy fix.

jshow:
I've been too busy to fix the conflict, if you have the time I'd appreciate it

Mostly waiting for comments and someone to merge it to dev - I'm using this branch in production presently

mhluska:
@jshow: https://github.com/Netflix/fast_jsonapi/pull/171

shishirmk:
Merged #171

rromanchuk:
This scope addition got me thinking about my design paradigms. I have many serializers that depend on my `rails` current_user, since this is just being added now i'm wondering if my solutions are poorly designed. 

Here's an example attribute that checks a simple hash if the current logged in user has viewed the object. There is no n+1 or required join, just an existence check of the logged in user. Let's say there is no scope support, how would you ever be able to accomplish this? 

```ruby
attribute :viewed_by_current_user do |object|
   current_user ? object.viewed_by?(current_user) : false
end
```

It might be bad habit from always having access to this scope from view templates

Issue #104, opened 2018-03-01
has_many relationship serialization


devtoro:
Hello everyone!

I recently started using fast_jsonapi in my rails API and I have to say that the performance improvement is amazing! 

I just have a suggestion to make. At the moment, in order to serialise "has_many" relationships, the gem is looking for **relationship_model_name**_ids and queries on that. With active record serializers, it would be enough to have a method to fetch the relationship data, named after the rails convention.

e.g.

Actor has_many :movies  could have a method _**movies**_ and the relationship data would be serialised. 

I mention this because I have some models that are Elasticsearch models and I was not storing the relationship ids, instead I had a method to fetch the relationship data ( I was caching the data as well ) and after I switched to fast_jsonapi the relationship was empty! :) 

For now, I store the ids in the mode, but I think It's an easy addition for the gem, if it makes sense to you as well.

shishirmk:
@devtoro we always look for **relationship_model_name_ids**(movie_ids) first as it is an optimization mechanism. It is common to just serialize only the ids unless the relationship is in the includes list. This saves unnecessary fetching of objects and object creation.

Looks like in your case you are saying if the **relationship_model_name_ids**(movie_ids)  method doesn't exist then instead of throwing an error or skipping serialization altogether, we should fallback on the **relationship_model_names(s)**(movies) method to get the id.

I think this request makes sense. Do you think you can take a shot at implementing it?

shuheiktgw:
PRed! https://github.com/Netflix/fast_jsonapi/pull/151

devtoro:
Sorry for the late reply...I had certain issues to attend and could not participate further. 

What's the status on this? 

Is [#151](https://github.com/Netflix/fast_jsonapi/pull/151) merged? Shall I take some action? I would gladly implement it if it's necessary. :)

PR #151, opened 2018-04-01
 Enable to fetch ids even when relationship_model_name_ids method is not implemented 

shuheiktgw:
This PR relates to https://github.com/Netflix/fast_jsonapi/issues/104.
I enabled fetch ids even when `id_method_name` is not implemented on the record. When the record raises `NoMethodError` by calling `id_method_name`, it tries to fetch ids via calling `record.public_send(object_method_name).map(&:id)`.

### DIFF
--- a/lib/fast_jsonapi/serialization_core.rb
+++ b/lib/fast_jsonapi/serialization_core.rb
@@ -42,7 +42,7 @@ module FastJsonapi
         polymorphic = relationship[:polymorphic]
 
         return ids_hash(
-          record.public_send(relationship[:id_method_name]),
+          fetch_ids(record, relationship[:id_method_name], relationship[:object_method_name]),
           relationship[:record_type]
         ) unless polymorphic
 
@@ -53,6 +53,13 @@ module FastJsonapi
         end if associated_object.respond_to? :map
 
         id_hash_from_record associated_object, polymorphic
+      end
+
+      def fetch_ids(record, id_method_name, object_method_name)
+        record.public_send(id_method_name)
+
+      rescue NoMethodError
+        record.public_send(object_method_name).map(&:id)
       end
 
       def attributes_hash(record)
--- a/lib/fast_jsonapi/serialization_core.rb
+++ b/lib/fast_jsonapi/serialization_core.rb
@@ -42,7 +42,7 @@ module FastJsonapi
         polymorphic = relationship[:polymorphic]
 
         return ids_hash(
-          fetch_ids(record, relationship[:id_method_name], relationship[:object_method_name]),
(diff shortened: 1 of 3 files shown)


lib/fast_jsonapi/serialization_core.rb
```diff
@@ -55,6 +55,13 @@ def ids_hash_from_record_and_relationship(record, relationship)
         id_hash_from_record associated_object, polymorphic
       end
 
+      def fetch_ids(record, id_method_name, object_method_name)
+        record.public_send(id_method_name)
+
+      rescue NoMethodError
+        record.public_send(object_method_name).map(&:id)
```
shishirmk: Looks like we are making an assumption that we can use `id` method to get the id of the object. if the object uses a different method to get id this would not work. Is there a way we can cover that case too?

shuheiktgw: Thank you so much for your feedback! You are absolutely right!
I assumed that I can get id by calling `id` method but it may not be that simple..

I have some idea how to solve the problem so let me work on it!

shuheiktgw: I added `id_from_origin` option to `has_many`, `has_one`, and `belongs_to`.
If you want to fetch id directly from the target object, all you need to do is simply set `id_from_origin: true`, and then `id_method_name` will be called upon the target objects.

I just want to see what you think about the idea so I haven't added tests or documents, but once it settles, I'll work on them! 👍 Oh, and also I will clean up the commits, too!

shishirmk: Need some more time to think about the repercussions of this approach. So might take some more time to get back to you

shuheiktgw: Sure, sure. Please let me know what you think whenever you have time! 👍

lib/fast_jsonapi/serialization_core.rb
```diff
@@ -55,6 +55,14 @@ def ids_hash_from_record_and_relationship(record, relationship)
         id_hash_from_record associated_object, polymorphic
       end
 
+      def fetch_ids(record, relationship)
+        if relationship[:id_from_origin]
```
shishirmk: I like this approach. I wonder if this if condition being called every time will have a noticeable performance impact. 

I know the performance tests need a refactor but is it possible to get a sense of the performance impact of this change?

shuheiktgw: @shishirmk Thank you so much for your comment! Let me see how this change affects the performance!

shishirmk:
Closing as we have no response

Issue #148, opened 2018-03-30
Way to pass in options for serialization


JasdipChauhan:
I'm looking to pass in conditions into a serializer to determine what kind of attribute it will return.

What I'm looking for: 
`UserSerializer.new(model, {use_token: true})`

and then in the serializer I can pull these options like so: 
`attribute :name do |object, options|
       return object.token if options[:use_token].present?
end`

I'm not sure if this piece of code in the docs already answers my question: 
`options = {}
options[:meta] = { total: 2 }
options[:include] = [:actors]
MovieSerializer.new([movie, movie], options).serialized_json
`
If it does, how do I access these custom options in the serializer class?

shishirmk:
@JasdipChauhan we don't have such a feature today. #139 also needs a similar feature. There is a pull request to address #139 in #150. You may want to bring up generalizing #150

shishirmk:
@christophersansone Any ideas on how we should implement this in the block custom attributes that we have today?

christophersansone:
@shishirmk Sorry I've been missing in action, been totally swamped, really want to do the refactor here soon.  This would be possible as designed, we would just have to store the entire `options` hash as an instance variable to allow user-defined keys to be specified and made available during the process, and then curry the options through to the `attributes_hash` method.

I'd probably like to first see a separate ticket for "specifying custom options when serializing" to make sure we design that properly.  Biggest concern would be preventing key collisions.  For example, the OP used this example:

```UserSerializer.new(model, {use_token: true})```

Here, `use_token` is a custom parameter, but it is defined the same way as the fixed serialization options such as `include`, and `meta`.  If FJ added a key in the future called `use_token`, it could have an adverse behavior here.  Likewise, if a user tried to add a custom option called `meta` or `include`, they may not know that those keys are reserved, and strange behavior can occur.  In general, we will want to define the fixed options and the custom options separately.  Either the custom options should be a nested hash of the options, or it could be a third parameter.

Issue #71, opened 2018-02-09
Proper separation between the serializer class vs. the serializer instance


christophersansone:
This issue is an extraction of the conversations in #49 and #67.

Right now, it is difficult to differentiate between what functionality belongs on the class and what belongs on the instance.  Currently they both kinda act as singletons.  It poses potential problems, currently and most notably when trying to implement AMS-style custom attribute methods, e.g.:

```ruby
class MovieSerializer
  include FastJsonapi::ObjectSerializer

  attributes :name, :year, :name_with_year

  def name_with_year
    "#{object.name} (#{object.year})"
  end
end
```

In what context is `name_with_year` executed?  What is its scope?  How is the `object` property available within this method?  And how do we avoid the performance pitfalls of AMS, that instantiates new serializers for every object being serialized?

I believe an ideal solution would be the following:

* The serializer **class** is responsible for containing the definition of the serializer: attributes, relationships, key transformation, etc.  Basically, anything built into the serializer DSL would be stored at the class.

* The serializer **instance** is responsible for serializing a single object at one time.  Its job is to take an object as an input and output the object's serialized representation.  This way, additional methods defined on the serializer (custom attributes, custom relationships, etc.) will naturally have an `object` within the context of the serialization process.

* The serializer **class** will have methods to serialize one object or a collection of objects.  The class will then create the necessary instance(s).  In most cases, developers will be calling class methods to perform the serialization, and they will not be instantiating serializers themselves.

* We can achieve performance levels comparable to today by reusing the serializer instances within the serialization process, as opposed to generating new instances for every object.  Prior to each iteration, we simply set the new `object` and clear any instance variables that custom methods may have set.  From the developer's point of view, it will feel like the serializer is new each time, but without the performance cost.

Examples:

```ruby
class MovieSerializer
  include FastJsonapi::ObjectSerializer

  attributes :name, :year, :name_with_year

  def name_with_year
    "#{object.name} (#{object.year})"
  end
end


MovieSerializer.to_hash(movie, options)
MovieSerializer.to_hash([ movie1, movie2 ], options)

MovieSerializer.to_json(movie, options)
MovieSerializer.to_json([ movie1, movie2 ], options)

```

Making some decisions along these lines will help set us up for success as we continue to build support for the entire JSON:API 1.0 spec.

shishirmk:
[...]

PR #153, opened 2018-04-04
Add in options[:scope] and receiving scope in attribute blocks

bbuck:
This feature is an alternative implementation to #150 that also should provide a solution for #139 and #145 and, potentially, #148.

Why I chose a different path: #150 opts to use a class-level `attr_accessor` for setting the scope, this approach feels open to potential race conditions or overlap as the scope may persist to other serializer instances or having more instances of one kind of serializer may unintentionally remove a scope -- basically the class-level approach feels open to unnecessary errors.

This solution instead opts for a local instance variable to the serializer instance and pass it upward to the class-level methods as necessary like it does for `@resource`.

**NOTE** There are some failing performance specs but they appear to be failing (at least for me) without my change. I wasn't able to notice any significant amount of decrease when my changes were present by if you do see any speed-ups please point them out!


jshow:
thank you @bbuck - I like this approach - I will review, good to get some eyes and thoughts on this

shishirmk:
@bbuck Thank you for working on this. I like this approach too. Do you see anything we could do to support #148 as well?

jshow:
@bbuck, to echo @christophersansone  comments in #148 , it may make sense to put the custom user options inside a standard subhash - `user_options` or something - to remove the chance of namespace collisions with keys used by fast_jsonapi

```
UserSerializer.new(model, {user_options: {scope: current_user, use_token: true})
```

bbuck:
@jshow Technically speaking that's what `scope` is. The point of the `:scope` key is to provide a context for the JSON to serialize in. In my example it's `current_user` but if your example:

```ruby
UserSerializer.new(model, {user_options: {scope: current_user, use_token: true})
```

You're scope should probably look more like:

```ruby
UserSerializer.new(model, {scope: {user: current_user, use_token: true})

# use
attribute :token do |user, scope|
  if scope[:use_token]
    user.get_token
  end
end
```

I'm not attached to the name `:scope` but it has clear meaning for it's purpose which is why I chose it. I don't see how renaming or nesting it will really change or fix anything unless scope becomes a portion of the JSON:API spec at some point (which seems unlikely).

bbuck:
@shishirmk I'm not sure if you've had a chance to review my comment above. The `scope` value is relatively open, and best practice is probably to always assign it a hash/struct-like object (so that future additions don't require rewriting the serializer). This can cover options, context variables, etc... required when serializing.

jdongelmans:
Any updates? I agree with @bbuck that every related issue can be solved using the ```scope``` parameter 👍

bbuck:
1) Totally agree, I was shooting for simplicity in initial examples but have since begun using a hash in the project in which I needed these changes too. Is the best approach for enforcing a hash to throw an error if it's not a hash?

2) I think so, but if the consensus is for it to change that's fine. What I see is exactly as you defined it, "the context in which code is being executed" as the scope contains data that defines the context in which the resource is being serialized. But I'm not going to hold up a feature for a name, if a better name is agreed on then it's easy to change.

christophersansone:
@bbuck Thanks!  To answer your questions:

1.  Yeah just throw an `ArgumentError` with a detailed message `unless params.is_a?(Hash)`.

2.  Agreed that my definition was lacking, and agreed that it's not that big of a deal either way, but the more precise we can make the term, the more easily it can be understood without a lot of explanation.  The difference between `scope` and `params` is subtle but I believe it's there.  I would describe the difference as this: `scope` contains the variables and methods _built into the context_, while `params` contain the variables and methods _passed into the context_.  The term `params` makes it clear that these values originated externally and were passed in, which is what we are doing here.  In the future, we intend to allow developers to define methods within the serializer class to calculate attribute values and such, and these would fall more into the category of scope.  [Suggested reading.](https://blog.codeship.com/a-deep-dive-into-ruby-scopes/) .

@shishirmk what do you think?  Anyone else have an opinion?

kyreeves:
First off, I'm loving this gem and thank you for all the work you all have done!

This PR would be very helpful, but I was wondering if there would be a way to use the scope (or params) to conditionally add attributes. You can see what I’m talking about [here in the AMS docs](https://github.com/rails-api/active_model_serializers/blob/v0.10.6/docs/general/serializers.md#attribute). I’m able to do something like this with #150 because I can access the scope on the class level, but after some trial and error, I’m having a hard time trying to implement it with this PR's approach.

christophersansone:
@kyreeves Thanks for the suggestion!  I'd say that is out of scope for this particular issue.  Here we are discussing enhancing _how_ the attribute values are determined, not determining _if_ the attributes should be included at all.  We can support a syntax similar to AMS to include an `if` proc, but it's outside the scope here.  Feel free to create a new issue!

kyreeves:
@christophersansone understood! Will do.

shishirmk:
@bbuck @christophersansone I like this approach over other approaches. I think we should allow for it to be generic and use the word `params` as @christophersansone suggested.

@bbuck once you make that change happy to merge it in.

bbuck:
@shishirmk Changes are in place!

mhluska:
Was there any reason for making params default to `nil` instead of `{}`? I'm finding I have to do `params && params[:foo]` in all my serializers now because it's not certain whether the caller will explicitly pass in a hash or not.

bbuck:
With the strict Hash enforcement now in place that probably makes more sense than `nil`. Does anyone else have an opinion.

bbuck:
cc @mhluska forgot to tag you

christophersansone:
@mhluska Good call!  Yes, now that `params` is a bona fide property, the developer should always be able to expect a hash.  It will feel similar to Rails controllers: `params` is always available, even if no request parameters were specified.

@bbuck Thanks for your work on this, you rock.  Would you mind making this adjustment?

The only other thing I think would be helpful is making `params` a public method.  It feels like a part of the public API at this point, and soon we will be adding the ability to define custom methods in the serializer for calculating attributes, relationships, etc.  @bbuck What do you think about adding an `attr_reader :params` to `ObjectSerializer`?

bbuck:
@christinaarmstrong With attributes being serialized at the class level, what benefit does an instance method to access params provide?

christophersansone:
@bbuck Probably nothing yet... I'm thinking more along the lines of #71 and being able to write custom methods within the serializer.  I'm not sure how useful it is outside of that, but in general it just seems like _a serializer has params_.

Take it a step further, and we can execute attribute blocks within the context of the serializer, giving us access to the `params` method without having to pass it in.  In other words, we can go from:

```ruby
attribute :viewed do |movie, params|
  params[:viewed_movies].include?(movie)
end
```

to:

```ruby
attribute :viewed do |movie|
  params[:viewed_movies].include?(movie)
end
```

Either way I'd say *mission accomplished* for this ticket, and we can choose to defer this until we have more reason to do so.

bbuck:
@christophersansone I'm not against any of those suggestions. My hesitations was with the current implementation not really requiring it. I think that's a great end goal to aspire too in simplifying the API for users (not having to know to provide a block requiring two arguments vs. one). I definitely liked that aspect of the #150, now #171.

christophersansone:
If a serializer instantiates other serializers, I would expect that ALL `options` would be passed down.  So yes, `params` should, and so should everything else... right?

mhluska:
I'm still having issues with the default params hash in some cases even after the latest changes here. But there's a possibility that it's caused by some cherry-picked PRs in my branch (I'm using the nested includes PR as well). Just thought I'd point it out.

bbuck:
@mhluska If you can be more specific I can potentially review and add test cases to verify that those concerns will be honored.

mhluska:
@bbuck in my case I have a `User` model with a `has_one :profile` in the serializer. In my user_controller show action, I do this:

```
render json: CurrentUserUserSerializer.new(
  User.includes(
    :profile,
    ...
    ).find(params[:id]),
    include: [:profile],
    params: { scope: current_user }
```

But then in the profile serializer, I try to rely on `params[:scope]` but it doesn't exist. My first thought is this is related to the `include` option but I haven't had the time to debug it.

lib/fast_jsonapi/object_serializer.rb
```diff
@@ -72,6 +72,8 @@ def process_options(options)
       @known_included_objects = {}
       @meta = options[:meta]
       @links = options[:links]
+      @params = options[:params] || {}
+      raise ArgumentError.new("`params` option passed to serializer must be a hash") unless @params.is_a?(Hash)
```
zinosama: A bit late to the party, but i'm curious why params has to be a hash?

bbuck: @zinosama Please review discussion starting here: https://github.com/Netflix/fast_jsonapi/pull/153#issuecomment-382211282

zinosama: Awesome. Reasonable. Thanks for the quick response 😄

bbuck: Yea no worries, sorry I did still link you to the wrong starter conversation but it seems you found the full convo!

zinosama: I opened https://github.com/Netflix/fast_jsonapi/pull/200 and meant to @ you but didn't find you in the @-list. It'd be reassuring if you could take a quick look and make sure i didn't miss anything?

theDisco:
Hey, thanks for implementing this very needed feature. I am waiting for it and currently installing the gem from the dev branch. Do you know when it is planned to release it?

arefaslani:
@shishirmk I really need this... When do you release this feature?

shishirmk:
@theDisco @arefaslani plan to release 1.2.0 this weekend

albertusnicholas:
Thank you so much we are currently use FastJson api for our project but we stuck with this feature that we need hope can use. Pleaseeeeeeeee :100:

dwiliandy:
Thank you so much, we need this feature for our project...

leonmojica:
Thank you very much for this fine piece of software, it sure makes my life a lot easier, any ideas when this will be released?

arefaslani:
> @theDisco @arefaslani plan to release 1.2.0 this weekend

PR #155, opened 2018-04-05
add object serializer inheritance of attributes, fixes #79

### LINKED ISSUE #79: Serializer Inheritance doesn't copy attributes
It's not clear that the following won't work as might be expected:

```
class UserSerializer

  attribute :first_name
  attribute :last_name
end
```
```
class FullUserSerializer < UserSerializer

  attribute :birthday
  attribute :email
end
```

Workaround:

```
class UserSerializer

  def self.inherited(subclass)
    subclass.attributes_to_serialize ||= {}
    subclass.attributes_to_serialize.merge!(attributes_to_serialize)
  end

  attribute :first_name
  attribute :last_name
end
```

eprothro:
This is to address issue #79.

Noteworthy is that this only addresses attribute inheritance (what is mentioned in that issue). It does not attempt to expand scope to relationship inheritance, etc. I'll have to leave that to someone else who has additional time.

### DIFF
--- a/lib/fast_jsonapi/object_serializer.rb
+++ b/lib/fast_jsonapi/object_serializer.rb
@@ -91,6 +91,12 @@ module FastJsonapi
     end
 
     class_methods do
+
+      def inherited(subclass)
+        super(subclass)
+        attributes_to_serialize.each { |name, method_or_block| subclass.attribute(name, method_or_block) }
+      end
+
       def reflected_record_type
         return @reflected_record_type if defined?(@reflected_record_type)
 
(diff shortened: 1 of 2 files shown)


shishirmk:
@eprothro Thanks for making the changes for the attributes inheritance. I think this is an important feature. I think having just attributes inheritance solved can cause confusion. Would really appreciate it if relationships, caching and other options where inherited as well.

eprothro:
Understand and agree! Unfortunately I don't have the time right now to dig into the relationship/caching/etc implementation to ensure I get it right. Someone with familiarity of the design should be able to add this functionality and test coverage very quickly. Apologies for the lack of availability at the moment.

shishirmk:
Taken care of by #203

PR #157, opened 2018-04-07
Fix serialization command in doc

albabar:
Rails generator names are case sensitive. In the README it was says `Serializer` but it should be `serializer`.


shishirmk:
This change is already in the dev branch will go to master with the next change

PR #168, opened 2018-04-13
Add scoping

juliantai:


shuheiktgw:
I think you should open PR to `dev` not to `master`! 👍

PR #169, opened 2018-04-15
Refactor serialization_core_spec

shuheiktgw:
I refactored and cleaned up the test in `serialization_core_spec.rb` with `subject` and `let`. Also I added some missing tests! Thank you! 👍

### DIFF
--- a/spec/lib/serialization_core_spec.rb
+++ b/spec/lib/serialization_core_spec.rb
@@ -1,82 +1,210 @@
 require 'spec_helper'
 
-describe FastJsonapi::ObjectSerializer do
-  include_context "movie class"
+describe FastJsonapi::SerializationCore do
+  include_context 'movie class'
   include_context 'group class'
 
-  context 'when testing class methods of serialization core' do
-    it 'returns correct hash when id_hash is called' do
-      inputs = [{id: 23, record_type: :movie}, {id: 'x', record_type: 'person'}]
-      inputs.each do |hash|
-        result_hash = MovieSerializer.send(:id_hash, hash[:id], hash[:record_type])
-        expect(result_hash[:id]).to eq hash[:id].to_s
-        expect(result_hash[:type]).to eq hash[:record_type]
-      end
+  describe '#self.id_hash' do
+    subject(:id_hash) { MovieSerializer.id_hash(id, 'movie') }
 
-      result_hash = MovieSerializer.send(:id_hash, nil, 'movie')
-      expect(result_hash).to be nil
-    end
+    context 'when id is string' do
+      let(:id) { 'x' }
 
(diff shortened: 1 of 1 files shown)


spec/lib/serialization_core_spec.rb
```diff
@@ -1,82 +1,212 @@
...
+        expect(id_hash[:type]).to eq 'movie'
       end
+    end
+
+    context 'when id is blank' do
+      let(:id) { nil }
 
-      result_hash = MovieSerializer.send(:id_hash, nil, 'movie')
-      expect(result_hash).to be nil
+      it 'returns nil' do
```
ankit8898: No need for `it 'returns nil'`,  `is_expected.to be_nil` will give the same message. Can use `it { is_expected.to be_nil }`

shuheiktgw:
@ankit8898 Thank you so much for your comment! I squashed and modify the test you mentioned above! 👍

shishirmk:
@ankit8898 The commits are squashed is it ok to merge?

spec/lib/serialization_core_spec.rb
```diff
@@ -1,82 +1,210 @@
...
+    end
+  end
+
+  describe '#self.ids_hash_from_record_and_relationship' do
+    context 'when an association is not polymorphic' do
+      subject(:ids_hash) { MovieSerializer.ids_hash_from_record_and_relationship(movie, relationship) }
+
+      let(:relationship) { { name: :actors, relationship_type: :has_many, object_method_name: :actors, id_method_name: :actor_ids, record_type: :actor } }
+
+      it 'returns the correct hash' do
```
ankit8898: This returns Array. Can we update example desc?

spec/lib/serialization_core_spec.rb
```diff
@@ -1,82 +1,210 @@
...
       end
     end
 
-    it 'returns the correct empty result when relationships_hash is called' do
-      movie.actor_ids = []
-      movie.owner_id = nil
-      relationships_hash = MovieSerializer.send(:relationships_hash, movie)
-      expect(relationships_hash[:actors][:data]).to eq([])
-      expect(relationships_hash[:owner][:data]).to eq(nil)
+    context 'when attribute without block' do
```
ankit8898: Should it be when attribute with block?

```ruby
context 'when attribute with block' do
 #...
end 
``` 
As you are passing a block to `MovieSerializer.attribute` ?

ankit8898:
@shishirmk I missed submitting my additional review, realized after coming back from vacation. 😄 

@shuheiktgw Can you have a look at the additional one's? Mainly around example descriptions. Thanks!

shishirmk:
Closing as we have no response

PR #174, opened 2018-04-18
Clean up multi_to_json

shuheiktgw:
I cleaned up `multi_to_json.rb` in the following two ways.

- Changed some unnecessarily `public` method private.
- Removed unused `self.reset_to_json!` method

### DIFF
--- a/lib/fast_jsonapi/multi_to_json.rb
+++ b/lib/fast_jsonapi/multi_to_json.rb
@@ -43,9 +43,20 @@ module FastJsonapi
       end
     end
 
-    def self.logger(device=nil)
-      return @logger = Logger.new(device) if device
-      @logger ||= Logger.new(IO::NULL)
+    def self.to_json(object)
+      _fast_to_json(object)
+    rescue NameError
+      define_to_json(FastJsonapi::MultiToJson)
+      _fast_to_json(object)
+    end
+
+    private
+
+    def self.define_to_json(receiver)
+      cl = caller_locations[0]
+      method_body = to_json_method
+      logger.debug { "Defining #{receiver}._fast_to_json as #{method_body.inspect}" }
+      receiver.instance_eval method_body, cl.absolute_path, cl.lineno
     end
 
     # Encoder-compatible with default MultiJSON adapters and defaults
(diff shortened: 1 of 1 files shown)


shuheiktgw:
This PR is meaningless if https://github.com/Netflix/fast_jsonapi/pull/177 is merged

shishirmk:
@shuheiktgw I like #177 lets see what @bf4 says about the changes. They look good to me

lib/fast_jsonapi/multi_to_json.rb
```diff
@@ -43,9 +43,20 @@ def rescue
...
-      return @logger = Logger.new(device) if device
-      @logger ||= Logger.new(IO::NULL)
+    def self.to_json(object)
+      _fast_to_json(object)
+    rescue NameError
+      define_to_json(FastJsonapi::MultiToJson)
+      _fast_to_json(object)
+    end
+
+    private
```
bf4: I think what you might mean here is to apply `# @api private` to the below methods, which I disagree with. It might make more sense to document intended usage, but I would say this is all a stable public API.

The private keyword has no effect on the interface.

In any case, the argument I can see for declaring it `# @api private` would be to say to anyone using the library not to rely on those methods being there, whether or not their public.  (In Ruby, nothing's really private anyhow).

lib/fast_jsonapi/multi_to_json.rb
```diff
@@ -43,9 +43,20 @@ def rescue
       end
     end
 
-    def self.logger(device=nil)
-      return @logger = Logger.new(device) if device
-      @logger ||= Logger.new(IO::NULL)
```
bf4: ❓

lib/fast_jsonapi/multi_to_json.rb
```diff
@@ -76,23 +87,9 @@ def self.to_json_method
...
-    def self.define_to_json(receiver)
-      cl = caller_locations[0]
-      method_body = to_json_method
-      logger.debug { "Defining #{receiver}._fast_to_json as #{method_body.inspect}" }
-      receiver.instance_eval method_body, cl.absolute_path, cl.lineno
-    end
-
-    def self.reset_to_json!
-      undef :_fast_to_json if method_defined?(:_fast_to_json)
-      logger.debug { "Undefining #{receiver}._fast_to_json" }
```
bf4: ❓ It's not used in the normal lifecycle of the app and because there's no test in the suite, but I wouldn't say it serves no purpose.  Especially if this module were made `module_function` and included.

I do have some tests I started writing in https://github.com/bf4/fast_multi_json but they didn't cover this specifically

shuheiktgw:
@bf4 Thank you so much for your comments! I should probably reconsider about this PR...

Issue #135, opened 2018-03-22
Regression: Date/Time is no longer serialized as ISO8601


martndemus:
I upgraded from 1.0.17 to 1.1.1 and somewhere (likely #58) the way date/timestamps are serialized regressed from being ISO8601 strings to simple time strings. Luckily my serializer specs caught it.

```diff
-      "created_at": "2018-03-22T09:45:00.246Z",
-      "updated_at": "2018-03-22T09:45:00.246Z",
+      "created_at": "2018-03-22 09:45:00 UTC"
+      "updated_at": "2018-03-22 09:45:00 UTC",
```

bf4:
The change

```diff
-        MultiJson.dump(payload) if payload.present?
+        FastJsonapi::MultiToJson.to_json(payload) if payload.present?
```

should used the same oj configs as multijson, as well as for the other json gems.  I'm guessing you don't have Oj installed and got bit by the bug fixed by https://github.com/Netflix/fast_jsonapi/pull/130

bf4:
@shishirmk IIRC, the bug fixed by #130 only affected installs lacking Oj. Perhaps I'm wrong about that. In any case it should be pretty easy to debug what's going on by 

```ruby
FastJsonapi::MultiToJson.to_json_method # => what's the output of this?

# and to remove any doubt
$VERBOSE = true
FastJsonapi::MultiToJson.logger(STDOUT)
FastJsonapi::MultiToJson.reset_to_json!
FastJsonapi::MultiToJson.to_json(payload)
```

Another possibility is that it's using  `%(::ActiveSupport::JSON.encode(object))` which wasn't part of the MultiJSON cascade.

It would be easier to reproduce and debug in discussion with @martndemus which is why I offered. There's some info that would be helpful just as other gems installed and a unix time that could be used for comparison.

martndemus:
Adding OJ as dep did fix it. So I guess that's what's happening. Would adding OJ by default make sense to keep backwards compat?

bf4:
@martndemus OJ depends on CRuby.  did you try debugging the code as I suggested? I think if you were to `bundle open fast_jsonapi` and comment out the active support line, you'd get your original behavior.

neongrau:
Having the same issue. But added to that when i installed OJ, serialization times doubled. 
[...]

PR #177, opened 2018-04-20
Make to_json twice as fast as it used to be

shuheiktgw:
I found the current implementation of `to_json` can be slow since it defines `to_json` method dynamically first time the method is called. So I use `class_eval` instead and change to define the method when class itself is evaluated. As a result `to_json` method bacame two times faster (from `0.4819999448955059` ms to `0.2079999540001154` ms) and the class became much simpler!

Below is the test code I used to check the performance.

```
require 'spec_helper'

describe FastJsonapi::MultiToJson do
  include_context 'movie class'

  before(:all) { GC.disable }
  after(:all) { GC.enable }

  describe 'self.to_json' do
    it do
      second = Benchmark.measure { FastJsonapi::MultiToJson.to_json movie }.real * 1000
      print 'to_json: '
      puts "#{second} ms"
    end
  end
end
```

### DIFF
--- a/lib/fast_jsonapi/multi_to_json.rb
+++ b/lib/fast_jsonapi/multi_to_json.rb
@@ -43,14 +43,9 @@ module FastJsonapi
       end
     end
 
-    def self.logger(device=nil)
-      return @logger = Logger.new(device) if device
-      @logger ||= Logger.new(IO::NULL)
-    end
-
     # Encoder-compatible with default MultiJSON adapters and defaults
     def self.to_json_method
-      encode_method = String.new(%(def _fast_to_json(object)\n ))
+      encode_method = String.new(%(def self.to_json(object)\n ))
       encode_method << Result.new(LoadError) {
         require 'oj'
         %(::Oj.dump(object, mode: :compat, time_format: :ruby, use_to_json: true))
@@ -76,23 +71,6 @@ module FastJsonapi
       encode_method << "\nend"
     end
 
-    def self.to_json(object)
-      _fast_to_json(object)
-    rescue NameError
-      define_to_json(FastJsonapi::MultiToJson)
-      _fast_to_json(object)
-    end
-
-    def self.define_to_json(receiver)
-      cl = caller_locations[0]
-      method_body = to_json_method
-      logger.debug { "Defining #{receiver}._fast_to_json as #{method_body.inspect}" }
-      receiver.instance_eval method_body, cl.absolute_path, cl.lineno
-    end
-
-    def self.reset_to_json!
-      undef :_fast_to_json if method_defined?(:_fast_to_json)
-      logger.debug { "Undefining #{receiver}._fast_to_json" }
-    end
+    class_eval to_json_method,__FILE__, __LINE__
   end
 end
--- a/spec/lib/multi_to_json_spec.rb
+++ b/spec/lib/multi_to_json_spec.rb
@@ -1,0 +1,11 @@
+require 'spec_helper'
+
+describe FastJsonapi::MultiToJson do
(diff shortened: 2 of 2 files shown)


shishirmk:
@shuheiktgw Thanks for the pull request @bf4 do you mind taking a look into this?

bf4:
@shishirmk I think there's a lot to discuss in this PR.

Firstly, is the argument that the *first* time the method is called is twice as fast?  If so, I don't think that's a useful benchmark. Mostly applications will live to have the method called more than that.

For what it's worth, this pattern comes from a perf PR to ActiveModelSerializers that still holds up https://github.com/rails-api/active_model_serializers/pull/233

The rest of the changes in this PR need a discussion of cost benefit analysis, including taking into account that the original PR was merged without a test suite, but was tested otherwise.  The various methods removed in this PR are critical to configuring and debugging the behavior.  I would use it in https://github.com/Netflix/fast_jsonapi/issues/135

bf4:
On my machine, a MBP running Ruby 2.4.1 I run the bench against master and get about 0.28ms each run and against make_to_json_fast about 0.11 ms each run.

But, if I change the benchmark to run to_json just once, I get the same 0.03 ms on each branch.  

```diff
require 'spec_helper'

describe FastJsonapi::MultiToJson do
  include_context 'movie class'

  before(:all) { GC.disable }
  after(:all) { GC.enable }

  describe 'self.to_json' do
    it do
+ FastJsonapi::MultiToJson.to_json movie
      second = Benchmark.measure { FastJsonapi::MultiToJson.to_json movie }.real * 1000
      print 'to_json: '
      puts "#{second} ms"
    end
  end
end
```

Let's rewrite the benchmark to use `Benchmark.bmbm`

> Sometimes benchmark results are skewed because code executed earlier encounters different garbage collection overheads than that run later. bmbm attempts to minimize this effect by running the tests twice, the first time as a rehearsal in order to get the runtime environment stable, the second time for real. GC.start is executed before the start of each of the real timings; the cost of this is not included in the timings. In reality, though, there’s only so much that bmbm can do, and the results are not guaranteed to be isolated from garbage collection and other effects.

```ruby
require 'spec_helper'

describe FastJsonapi::MultiToJson do
  include_context 'movie class'

  describe 'self.to_json' do
    it do
      n = 1_000
      Benchmark.bmbm do |x|
        x.report('to_json') {
          i = 0
          while i < n
            FastJsonapi::MultiToJson.to_json movie
            i += 1
          end
        }
      end
    end
  end
end
```

Here are the outputs against each branch.  I'm not going to label which branch is which 'cause they're the same :)

```
Rehearsal -------------------------------------------
to_json   0.020000   0.000000   0.020000 (  0.032862)
---------------------------------- total: 0.020000sec

              user     system      total        real
to_json   0.020000   0.000000   0.020000 (  0.021913)
```.

```
📝 $ be rspec spec/bench_spec.rb
Rehearsal -------------------------------------------
to_json   0.030000   0.010000   0.040000 (  0.029570)
---------------------------------- total: 0.040000sec

              user     system      total        real
to_json   0.020000   0.000000   0.020000 (  0.022821)
```

bf4:
@shuheiktgw I commented above

shuheiktgw:
@bf4 Thank you so much for your comments! You are absolutely right if you take a benchmark multiple time the result will be the same since the difference appears only for the first time that `to_json` is called. 

Even if this PR does not improve performance, I personally believe that I can simply `to_json` method by using `class_eval`. What would you think if I make back some logging features I removed?

bf4:
@shuheiktgw Given this discussion, I think you should rename the PR to describe what you want to do with it.

PR #178, opened 2018-04-20
Use proper casing on generator example

FanaHOVA:
`rails g Serializer` isn't a valid command, serializer shouldn't be capitalized.


ankit8898:
Thank you! 👍

PR #186, opened 2018-04-24
Refactor tests for key_transform method

shuheiktgw:
I refactored tests for `key_transform` using shared_example and added some missing tests!
Thank you! 👍


spec/lib/object_serializer_class_methods_spec.rb
```diff
@@ -155,10 +151,70 @@
         end
       end
 
+      after do
+        MovieSerializer.attributes_to_serialize.delete(:title_with_year)
+      end
```
shuheiktgw: I found a small bug in my previous commit. I should not do `MovieSerializer.attributes_to_serialize = {}`. It will mistakenly cleaned up all the attributes from MovieSerializer.

Issue #73, opened 2018-02-09
Make `belongs_to` an alias of `has_one`


vovimayhem:
Looking at the code, there's no real difference for an association being `belongs_to` and  `has_one`. Instead, is a source for problems (two places with the same code are two places you can forget to place a fix.. have you noticed the [default serializer is obtained differently?](https://github.com/Netflix/fast_jsonapi/blob/dev/lib/fast_jsonapi/object_serializer.rb#L188)). Both do the same thing, and both cause the same consequences in the serialization core.

Also, the jsonapi doesn't care about the direction of a 1-to-1 association.

How about:

* Removing the [`belongs_to`](https://github.com/Netflix/fast_jsonapi/blob/dev/lib/fast_jsonapi/object_serializer.rb#L186) method in `object_serializer.rb`
* Adding `belongs_to` as an alias of `has_one`, so it is backwards-compatible.

shishirmk:
@vovimayhem That is a very nice way to refactor some of the repetition we have in the code today. You are right as far JSON:API is concerned i don't think there is any distinction between belongs_to and has_one. I think we should work on this idea.

amrrbakry:
I think I can help with this. I will open a PR soon :+1:

vovimayhem:
Hey
I've got a commit with that on another PR (Sorry!)

vovimayhem:
c5295e6

amrrbakry:
No problem. Good work 👍

shishirmk:
This should be released in version 1.2 of the gem

ababich:
I think this is a big problem when `has_one` and `belongs_to` are aliased they are logically different and it should be implemented:

A `belongs_to` B means A has `b_id` attribute
A `has_one` B means B has `a_id` attribute

Current implementation when you have to provide additional attribute for `has_one` relation is pretty inconvenient

abevoelker:
I agree with @ababich and created #257 for consideration

vovimayhem:
@ababich @abevoelker IMHO the direction/cardinality of the association in the database shouldn't be relevant on the serialization part of JSONAPI.

In both cases, you get the associated object's `id` and `type`, regardless if the ActiveRecord/ORM model your'e serializing has it implemented as a `belongs_to` or a `has_one`. There should be no extra attributes/options required to declare on the serializer.

ababich:
@vovimayhem 

[...]

PR #188, opened 2018-04-25
DRY up some code

oniofchaos:
Found that this code was basically repeated three times. Condensed it up into a single method.

### DIFF
--- a/lib/fast_jsonapi/object_serializer.rb
+++ b/lib/fast_jsonapi/object_serializer.rb
@@ -169,32 +169,26 @@ module FastJsonapi
       def has_many(relationship_name, options = {})
         name = relationship_name.to_sym
         singular_name = relationship_name.to_s.singularize
-        add_relationship(
-          name,
-          create_relationship_hash(
-            relationship_name, name, options, singular_name, :has_many, '_ids'
-          )
+        hash = create_relationship_hash(
+          relationship_name, name, options, singular_name, :has_many, '_ids'
         )
+        add_relationship(name, hash)
       end
 
       def belongs_to(relationship_name, options = {})
         name = relationship_name.to_sym
-        add_relationship(
-          name,
-          create_relationship_hash(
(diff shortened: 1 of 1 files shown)


lib/fast_jsonapi/object_serializer.rb
```diff
@@ -169,57 +169,48 @@ def add_relationship(name, relationship)
...
+        add_relationship(
+          name,
+          create_relationship_hash(
+            relationship_name, name, options, relationship_name, :has_one
+          )
+        )
+      end
+
+      def create_relationship_hash(
+        base_key, name, options, base_serialization_key, relationship_type, id_postfix='_id'
```
shishirmk: really big list of params is there anyway we can reduce this?

lib/fast_jsonapi/object_serializer.rb
```diff
@@ -169,57 +169,48 @@ def add_relationship(name, relationship)
...
-          record_type: record_type,
-          object_method_name: options[:object_method_name] || name,
-          serializer: compute_serializer_name(serializer_key),
-          relationship_type: :belongs_to,
-          cached: options[:cached] || true,
-          polymorphic: fetch_polymorphic_option(options)
-        })
+        add_relationship(
+          name,
+          create_relationship_hash(
```
shishirmk: it might be clearer and explicit to compute the relationship_hash first and then pass it to `add_relationship`

```ruby
relationship_hash = create_relationship_hash(arg1, arg2, arg3)
add_relationship(name, relationship_hash)
```

lib/fast_jsonapi/object_serializer.rb
oniofchaos: Got it down to 3 params

oniofchaos:
Did some cleanup, let me know what you think.

oniofchaos:
Also I added a change to address https://github.com/Netflix/fast_jsonapi/issues/73

Issue #165, opened 2018-04-12
Proc for association


mhluska:
Is there any way to provide a proc for a `belongs_to` association? In `active_model_serializers`, I do:

```
belongs_to :buyer do
  object.payable.buyer
end
```

In other words, I have a belongs to relationship through another model. In my case video belongs to sale belongs to buyer. But I want my client to think video belongs to sale. 

Is there any way to do this in `fast_jsonapi`? As far as I can tell procs can only be used for attributes and not associations.

shishirmk:
Both these are enhancements we need to add. On second question there has been some progress and I am leaning towards merging #153

mhluska:
@shishirmk what is the current way to hide relationships based on authorization? It seems to me that theres no way to actually use this in production right now because I can't hide sensitive data from an unauthorized user (for example, the list of videos a user bought in my case).

shuheiktgw:
I just started to work on the first problem, enabling to use block to define relationship.

PR #195, opened 2018-04-30
Enable to use block to define relationship

shuheiktgw:
This PR relates to https://github.com/Netflix/fast_jsonapi/issues/165.

I enabled to use block to define relationships. For example, if `An actor belongs to an agency, and an agency belongs to a city`, you can refer to the city directly from `ActorSerializer` like below.

```
class ActorSerializer
  attributes :name, :email

  belongs_to :city do |actor|
    actor.agency.city
  end
end
```

Thank you! 👍

### DIFF
--- a/lib/fast_jsonapi/serialization_core.rb
+++ b/lib/fast_jsonapi/serialization_core.rb
@@ -48,11 +48,11 @@ module FastJsonapi
         polymorphic = relationship[:polymorphic]
 
         return ids_hash(
-          record.public_send(relationship[:id_method_name]),
+          fetch_id(record, relationship),
           relationship[:record_type]
         ) unless polymorphic
 
-        return unless associated_object = record.send(relationship[:object_method_name])
+        return unless associated_object = fetch_associated_object(record, relationship)
 
         return associated_object.map do |object|
           id_hash_from_record object, polymorphic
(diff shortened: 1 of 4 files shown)


shishirmk:
We might need to add params support. Just the same way they have been added to attributes here
https://github.com/Netflix/fast_jsonapi/tree/dev#params

shishirmk:
Will merge this in and then add the params part myself

lib/fast_jsonapi/serialization_core.rb
```diff
@@ -130,6 +128,22 @@ def get_included_records(record, includes_list, known_included_objects)
...
+        return relationship[:object_block].call(record) unless relationship[:object_block].nil?
+        record.send(relationship[:object_method_name])
+      end
+
+      def fetch_id(record, relationship)
+        unless relationship[:object_block].nil?
+          object = relationship[:object_block].call(record)
+
+          return object.map(&:id) if object.respond_to? :map
+          return object.id
```
mhluska: @shishirmk is this designed to work that returns `nil`? If I return `nil` in the block, this throws a runtime error.

shishirmk: Thats something we would have to fix.

PR #198, opened 2018-04-30
Use the serializer_key directly if it is not a symbol in relationships

carsol:
### Context:
This allows us to explicitly set the Serializer class in relationships without automagically adding the original model's namespace.

For example:
```
module Foo
  class FooItemSerializer < ApplicationSerializer

    attributes :description, :name

    has_many :bars, serializer: CoolApp::Api::V1::BarSerializer
  end
end
```

Without this change, it would throw an exception ```"uninitialized constant Foo::CoolApp::Api::V1::BarSerializer"``` because it's appending the parent namespace and can't find that serializer.

If you're passing in a symbol (default behavior) then it would classify and append the namespace correct as before.

Let me know if you have any questions.

---
This change is [Reviewable](https://reviewable.io/reviews/netflix/fast_jsonapi/198)


shishirmk:
@carsol i am trying to merge it to dev. Could you look into the conflicts?

carsol:
@shishirmk I just merged in `dev` so it should be up to date now!

PR #200, opened 2018-05-02
Remove duplicate id_hash call from #record_hash

zinosama:
A duplicate call to #id_hash seems to have got in during rebase from https://github.com/Netflix/fast_jsonapi/pull/153/files#diff-0d22ea72f1e72c75e7887cdbcba46871R89


zinosama:
Spec failure is unrelated and fixed by #201

bbuck:
Great catch! I must have missed that resolving the merge conflicts.

PR #201, opened 2018-05-02
Use string for serializer name

zinosama:
This PR fixes the a spec broken by https://github.com/Netflix/fast_jsonapi/pull/198.


zinosama:
@carsol

carsol:
Nice! Thank you for this @zinosama.

PR #206, opened 2018-05-09
Always return a symbol from compute_serializer_name.

carsol:
I made a previous change that if you passed a class instead of a symbol, it would just return it (https://github.com/Netflix/fast_jsonapi/pull/198). This method works either way but it makes more sense to be uniform about what we return (always a symbol class). It also allows for us to clean up a test also.

### Related PRs:
#198
#201

---
This change is [Reviewable](https://reviewable.io/reviews/netflix/fast_jsonapi/206)

### DIFF
--- a/lib/fast_jsonapi/object_serializer.rb
+++ b/lib/fast_jsonapi/object_serializer.rb
@@ -221,7 +221,7 @@ module FastJsonapi
       end
 
       def compute_serializer_name(serializer_key)
-        return serializer_key unless serializer_key.is_a? Symbol
+        return serializer_key.to_s.to_sym unless serializer_key.is_a? Symbol
         namespace = self.name.gsub(/()?\w+Serializer$/, '')
         serializer_name = serializer_key.to_s.classify + 'Serializer'
         return (namespace + serializer_name).to_sym if namespace.present?
--- a/spec/lib/object_serializer_class_methods_spec.rb
+++ b/spec/lib/object_serializer_class_methods_spec.rb
@@ -251,7 +251,7 @@ describe FastJsonapi::ObjectSerializer do
           attributes :name, :release_year
           has_many :actors
           belongs_to :owner, record_type: :user
-          belongs_to :movie_type, serializer: "#{key_transform}_movie_type".to_sym
+          belongs_to :movie_type, serializer: "#{key_transform}_movie_type"
         end
         movie_type_serializer_class = Object.const_set(movie_type_serializer_name, Class.new)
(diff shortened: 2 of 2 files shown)


carsol:
Related PRs: 
- https://github.com/Netflix/fast_jsonapi/pull/198
- https://github.com/Netflix/fast_jsonapi/pull/201

lib/fast_jsonapi/object_serializer.rb
```diff
@@ -221,7 +221,7 @@ def create_relationship_hash(base_key, relationship_type, options, block)
       end
 
       def compute_serializer_name(serializer_key)
-        return serializer_key unless serializer_key.is_a? Symbol
+        return serializer_key.to_s.to_sym unless serializer_key.is_a? Symbol
```
zinosama: The output of this method is always used as
`@relationships_to_serialize[item][:serializer].to_s.constantize` 

so why not change this method to simply return a string? Not sure what's the point of converting to symbol then back to string.

carsol: Good point. I think it would just make sense always return a `constantize` serializer Class instead of a String or Symbol. It cuts the conversation.

So this would look like this:
```
      def compute_serializer_name(serializer_key)
        return serializer_key.to_s.constantize unless serializer_key.is_a? Symbol
        namespace = self.name.gsub(/()?\w+Serializer$/, '')
        serializer_name = serializer_key.to_s.classify + 'Serializer'
        return (namespace + serializer_name).constantize if namespace.present?
        (serializer_key.to_s.classify + 'Serializer').constantize
      end
```

By doing this we can make the output be:
```
@relationships_to_serialize[item][:serializer]
```

Thoughts @zinosama ?

zinosama: If the serializer class doesn't exist, this approach will display a more accurate/upstream source of error - thus I think it's a better solution.

spec/lib/object_serializer_class_methods_spec.rb
```diff
@@ -251,7 +251,7 @@
           attributes :name, :release_year
           has_many :actors
           belongs_to :owner, record_type: :user
-          belongs_to :movie_type, serializer: "#{key_transform}_movie_type".to_sym
+          belongs_to :movie_type, serializer: "#{key_transform}_movie_type"
```
zinosama: Also, i'm not sure what's the reason for this change - it seems to fail the spec 😞

PR #210, opened 2018-05-14
[fix] Optional belongs_to

malina:
If belongs_to relation was empty there was the error _undefined method `id' for nil:NilClass_

### DIFF
--- a/lib/fast_jsonapi/serialization_core.rb
+++ b/lib/fast_jsonapi/serialization_core.rb
@@ -149,6 +149,7 @@ module FastJsonapi
                 included_records.concat(serializer_records) unless serializer_records.empty?
               end
 
+              next unless inc_obj
               code = "#{record_type}_#{inc_obj.id}"
               next if known_included_objects.key?(code)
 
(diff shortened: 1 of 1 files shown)


lib/fast_jsonapi/serialization_core.rb
```diff
@@ -149,6 +149,7 @@ def get_included_records(record, includes_list, known_included_objects, params =
                 included_records.concat(serializer_records) unless serializer_records.empty?
               end
 
+              next unless inc_obj
```
llenodo: I ran into the same error. Seems like they're trying to solve the issue above on line 144:
```
included_objects = [included_objects] unless relationship_type == :has_many
next if included_objects.blank?
```

The problem is that included objects is an array containing an emtpy element `[nil]`

I think if we just switch the order of the above 2 methods it will solve the problem and then you dont need to add this nil check

```
next if included_objects.blank?
included_objects = [included_objects] unless relationship_type == :has_many
```

shishirmk:
@malina we should write a test to demonstrate this issue. I agree with @llenodo i think the issue would be fixed by just changing the order of those 2 lines.

llenodo:
I opened up https://github.com/Netflix/fast_jsonapi/pull/216 to fix this issue. Issue filed here https://github.com/Netflix/fast_jsonapi/issues/215

PR #218, opened 2018-05-20
making object level links similar to attributes

shishirmk:


shishirmk:
@jshow is there a way to add access to route helpers in the serializer?

jshow:
@shishirmk  the solution I coded gave ran the block within the scope of the instance of the serializer

giving access to all injected methods

shishirmk:
@jshow true it did. However it's confusing to have access to the instance inside a class method. 

Can we add the route helpers as class methods for now? I realize that the experience is not great but this was an unintended consequence of some initial design decisions.

We can revisit this in may be the next major version

PR #234, opened 2018-06-03
Use size to determine if resource is a collection

### LINKED ISSUE #229: Sequel gem compatibility
If an app is using the Sequel gem as their ORM adapter, a model instance will have `each` defined on it, for the purpose of iterating all values on that particular instance.  

This will prevent the serialization of model instances when `FastJsonapi::ObjectSerializer` calls `is_collection?` [here](https://github.com/Netflix/fast_jsonapi/blob/master/lib/fast_jsonapi/object_serializer.rb#L84), from the `serializable_hash` method.

    def is_collection?(resource)
      resource.respond_to?(:each) && !resource.respond_to?(:each_pair)
    end

Perhaps `size` would also achieve the same effect here? 

`resource.respond_to?(:size) && !resource.respond_to?(:each_pair)`

trueinviso:
This PR fixes #229.

### DIFF
--- a/lib/fast_jsonapi/object_serializer.rb
+++ b/lib/fast_jsonapi/object_serializer.rb
@@ -82,7 +82,7 @@ module FastJsonapi
     end
 
     def is_collection?(resource)
-      resource.respond_to?(:each) && !resource.respond_to?(:each_pair)
+      resource.respond_to?(:size) && !resource.respond_to?(:each_pair)
     end
 
     class_methods do


ababich:
The fix works well for AR and Sequel but is not future-proof for others, I suggest to add ability for specific DB toolkits with fallback to some generic solution

ababich:
And, of course, additional option to force smth like `is_collection: true` via options is required

ababich:
Please, see https://github.com/Netflix/fast_jsonapi/pull/239 related to this

shishirmk:
#239 is more flexible

PR #235, opened 2018-06-05
remove accidentally duplicated line

panckreous:


shishirmk:
This has already been taken care of https://github.com/Netflix/fast_jsonapi/pull/232/files

Issue #175, opened 2018-04-19
Automatic lookup within Rails controllers


oniofchaos:
If I do something like `render json: model`, when I'm using AMS it automatically detects `ModelSerializer` but when I'm switching over to `fast_jsonapi` it doesn't seem to have that same magic. Am I messing something up or is that not a feature of this gem?

P.S. Overall love the gem and the philosophy, thanks for open sourcing this.

shishirmk:
Yes this is not automatic yet. However #89 and other pull requests have made it a bit less verbose.

This should work

```ruby
render json: WidgetSerializer.new(widget)
```

guilleiguaran:
I have a PR to add this ready but I'm a bit concerned about causing conflicts with AMS (I imagine that there are projects using both libraries simultaneously)

WDYT about using:

 ```ruby
render jsonapi: model
```

To avoid conflicts with `json` renderer overrided by AMS?

oniofchaos:
That doesn't seem necessary, as I feel like it should use the model name to load the appropriate serializer and that serializer would define whether it's using AMS or fast_jsonapi via what it's inheriting from/what it includes.

One thought - are you're saying that people have a model being serialized by both gems in different controllers/actions?

guilleiguaran:
@oniofchaos AMS right now is overriding ActionController::Renderers internal private methods making a bit hard your proposal. 

To use `:json` and keep the compatibility with AMS we will need to detect if AMS is in the project and override again the methods that it's overriding to make them aware of fast_jsonapi serializers and have another code path to override the default JSON renderer if AMS is not present in the project. It can be done but it'll be "hacky" in my opinion.

> One thought - are you're saying that people have a model being serialized by both gems in different controllers/actions?

I'm talking about people using AMS for some models and fast_jsonapi for other models, I think that is the case for some people looking to improving the performance in some serializers without having to migrate a whole existing project to fast_jsonapi.

stas:
I started a little project to help integrate `fast_jsonapi` with rails/<other frameworks>.

To be honest, I'm not sure how much of the JSONAPI spec `fast_jsonapi` want's to include. While working on the renderer side it's kinda clear that it's not just the serializer lookup support that needs to be provided, it's also the error handling/rendering support and the error serializers (#102) that should available.

Consider this, as an example implementation:
https://github.com/stas/jsonapi.rb/blob/master/lib/jsonapi/rails.rb#L28-L100

PR #243, opened 2018-06-09
Use a Railtie to extend Rails

guilleiguaran:
Railties are the core of the Rails framework and provide several hooks to extend Rails, using Railties is the 'Rails-way' to introduce extensions/patches to Rails libraries/components.

Based on this I'm going to add hooks to introduce a new renderer for Action Controller as described in https://github.com/Netflix/fast_jsonapi/issues/175

### DIFF
--- a/lib/extensions/has_one.rb
+++ b/lib/extensions/has_one.rb
@@ -1,20 +1,18 @@
 # frozen_string_literal: true
 
-if defined?(::ActiveRecord)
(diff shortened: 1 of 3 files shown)


ansonhoyt:
@guilleiguaran curious if you're still following this up with those hooks? From your comments on #175 and #89, this sound promising for smooth integration in controllers.

PR #249, opened 2018-06-14
Set type value when setting key transform

### LINKED ISSUE #248: Key transforms not called on the record type of included objects
When key_transform is set to anything other than `underscore` and `set_type` is not manually called and includes are requested then the includes object types are not transformed as expected.  

Looking in the code I see that `set_type` is called via inclusion into the class and even attempts to use the `run_key_transform` lookup, but given the code load/run order it will never be transformed.  You can call `set_type` manually after setting `key_transform` and it will be transformed as expected, but given that this is optional it should not be required for `key_transform` to function properly and it definitely shouldn't depend on the calling order of `key_transform` and `set_type`.

djohnson:
This simply adds a call to the `set_type` method from `set_key_transform`, which will ensure that the `record_type` is transformed as is expected.

This fixes #248

### DIFF
--- a/lib/fast_jsonapi/object_serializer.rb
+++ b/lib/fast_jsonapi/object_serializer.rb
@@ -118,6 +118,9 @@ module FastJsonapi
           underscore: :underscore
         }
         self.transform_method = mapping[transform_name.to_sym]
+
+        # ensure that the record type is correctly transformed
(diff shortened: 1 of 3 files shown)


shishirmk:
looks like you already have a pull request ready for this issue. I will review it this weekend and get back to you

PR #254, opened 2018-06-21
Adds the option to select the serializer attributes

g13ydson:
Adds the option to select the serializer attributes
This is important so that we do not have to create another serializer to return specific attributes. We can also select the attributes of relationships

```ruby
options = {}
options[:include] = [:actors]
options[:attributes] = {
  movie:  [:name],
  actors: [:email]
}
```

Replaces `Hash#fetch+args` bacause it's slower than `Hash#fetch+blocks`

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,9 +230,9 @@ class MovieSerializer
 end
 ```
 
-### Select attributes
+### Select fields
 
-Support for explicit attributes declaration through ` options[:attributes] `.
+Support for explicit fields declaration through ` options[:fields] `.
 
 ```ruby
 options = {}
--- a/README.md
+++ b/README.md
@@ -230,6 +230,16 @@ class MovieSerializer
 end
 ```
 
+### Select attributes
+
+Support for explicit attributes declaration through ` options[:attributes] `.
+
+```ruby
+options = {}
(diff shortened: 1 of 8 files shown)


shishirmk:
@g13ydson this is similar to #63. The only thing missing is the ability to chose the relationships that the user would like to serialize. It would be great if you could add the feature. Then i can review it and we can merge it in.

g13ydson:
Yes, it is possible to serialize the attributes of relationships. We are testing this modification and it is working perfectly.

ababich:
I believe this should be `fields` only as per JSONAPI spec. and this is duplicate of #63 

PS. I don't talk about implementation, haven't compare but I'm pretty sure API here is incorrect

g13ydson:
@ababich True, I thought about renaming attributes to fields.
the idea is similar, but PR #63 did not implement some things.

shishirmk:
@g13ydson how would i exclude a entire relationship from being serialized? For example what if i wanted to exclude all the actors relationship from the serialized movie object? That is allowed in the JSON API spec and #63 also had not fully implemented this feature.

shishirmk:
#271 implemented this feature so closing this pull request

PR #256, opened 2018-06-21
Conditional attributes

TrevorHinesley:
_This is mostly code from #66 by @andyjeffries, but the changes were cherry picked and put on top of the current dev branch to fix some tests that were failing._

```ruby
class MovieSerializer

  attributes :name, :year
  attribute :release_year, if: Proc.new do |record|
    # Release year will only be serialized if it's greater than 1990
    record.release_year > 1990
  end

  attribute :director, if: Proc.new do |record, params|
    # The director will be serialized only if the :admin key of params is true
    params && params[:admin] == true
  end
end

# ...
current_user = User.find(cookies[:current_user_id])
serializer = MovieSerializer.new(movie, { params: { admin: current_user.admin? }})
serializer.serializable_hash
```

### DIFF
--- a/lib/fast_jsonapi/object_serializer.rb
+++ b/lib/fast_jsonapi/object_serializer.rb
@@ -149,11 +149,18 @@ module FastJsonapi
 
       def attributes(*attributes_list, &block)
         attributes_list = attributes_list.first if attributes_list.first.class.is_a?(Array)
+        options = attributes_list.last.is_a?(Hash) ? attributes_list.pop : {}
         self.attributes_to_serialize = {} if self.attributes_to_serialize.nil?
+        self.optional_attributes_to_serialize = {} if self.optional_attributes_to_serialize.nil?
+
         attributes_list.each do |attr_name|
           method_name = attr_name
           key = run_key_transform(method_name)
-          attributes_to_serialize[key] = block || method_name
+          if options[:if].present?
+            optional_attributes_to_serialize[key] = [method_name, options[:if]]
+          else
+            attributes_to_serialize[key] = block || method_name
+          end
         end
       end
 
--- a/lib/fast_jsonapi/object_serializer.rb
+++ b/lib/fast_jsonapi/object_serializer.rb
@@ -3,6 +3,7 @@
 require 'active_support/core_ext/object'
 require 'active_support/concern'
 require 'active_support/inflector'
+require 'fast_jsonapi/attribute_serializer'
 require 'fast_jsonapi/serialization_core'
 
 module FastJsonapi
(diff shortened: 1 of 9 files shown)


TrevorHinesley:
@christophersansone I *believe* I accomplished what we were going for [here](https://github.com/Netflix/fast_jsonapi/pull/66#discussion_r196787603). Mind taking a look?

TrevorHinesley:
Also, is there any chance that the instantiation of `Attribute` could slow things down? Obviously, the performance tests are passing, but what if there's tons of attributes? I'm ignorant to whether the instantiation/GC issues in AMS that this library solves are related to actual class instantiation, or how it serialized instantiated classes in general (in other words, whether using instantiated classes like this will affect performance at all).

lib/fast_jsonapi/object_serializer.rb
```diff
@@ -156,7 +156,7 @@ def attributes(*attributes_list, &block)
         attributes_list.each do |attr_name|
           method_name = attr_name
           key = run_key_transform(method_name)
-          attributes_to_serialize[key] = AttributeSerializer.new(
+          attributes_to_serialize[key] = Attribute.new(
```
TrevorHinesley: Also, is there any chance that the instantiation of Attribute could slow things down? Obviously, the performance tests are passing, but what if there's tons of attributes? I'm ignorant to whether the instantiation/GC issues in AMS (issues that this library aims to solve) are related to actual class instantiation, or just how it serialized instantiated classes in general (in other words, whether using instantiated classes like this will affect performance at all).

shishirmk:
@TrevorHinesley Thank you very much. This looks neat. @christophersansone thanks for all the comments and help

andyjeffries:
Hey @shishirmk do you have any idea of when this will be released as part of the gem?

christophersansone:
@andyjeffries Thanks for getting the ball rolling and to keep nudging it along!

@shishirmk Thanks for the quick merge!

@TrevorHinesley Thanks for getting this over the finish line!  Quick response about performance.  Typically, any feature we add to this gem will reduce performance.  The fastest code is no code.  So the goal here is to offer the features needed while minimizing the performance impact.  I have not run any A/B performance tests here (anyone up for it?), but by reading the code it looks like we achieved it.  The real performance issues occur when the serializer has to do a lot of extra work upon each record serialization.  Here, we are creating the `Attribute` instances once at load time and referencing them for each serialization, which is ideal.  From there it's just a small additional code path to do the conditional check.  So I think we are in good shape here.  Even if there are thousands of attributes, it looks to be about as fast as it can be under those circumstances.

groony:
Great feature. I'm extremely waiting it in gem

PR #260, opened 2018-06-26
Conditional relationships

kyreeves:
Just like conditional attributes, conditional relationships can be defined by passing a Proc to the `if` key.

```
class MovieSerializer

  # Actors will only be serialized if the record has any associated actors
  has_many :actors, if: Proc.new { |record| record.actors.any? }

  # Owner will only be serialized if the :admin key of params is true
  belongs_to :owner, if: Proc.new { |record, params| params && params[:admin] == true }
end

# ...
current_user = User.find(cookies[:current_user_id])
serializer = MovieSerializer.new(movie, { params: { admin: current_user.admin? }})
serializer.serializable_hash
```

### DIFF
--- a/lib/fast_jsonapi/serialization_core.rb
+++ b/lib/fast_jsonapi/serialization_core.rb
@@ -146,6 +146,8 @@ module FastJsonapi
           items = parse_include_item(include_item)
           items.each do |item|
             next unless relationships_to_serialize && relationships_to_serialize[item]
+            conditional_proc = relationships_to_serialize[item][:conditional_proc]
+            next if conditional_proc && !conditional_proc.call(record, params)
             raise NotImplementedError if @relationships_to_serialize[item][:polymorphic].is_a?(Hash)
             record_type = @relationships_to_serialize[item][:record_type]
             serializer = @relationships_to_serialize[item][:serializer].to_s.constantize
--- a/lib/fast_jsonapi/serialization_core.rb
+++ b/lib/fast_jsonapi/serialization_core.rb
@@ -34,47 +34,15 @@ module FastJsonapi
         end
       end
 
-      def ids_hash(ids, record_type)
-        return ids.map { |id| id_hash(id, record_type) } if ids.respond_to? :map
-        id_hash(ids, record_type) # ids variable is just a single id here
-      end
-
-      def id_hash_from_record(record, record_types)
-        # memoize the record type within the record_types dictionary, then assigning to record_type:
-        record_type = record_types[record.class] ||= record.class.name.underscore.to_sym
-        id_hash(record.id, record_type)
-      end
-
-      def ids_hash_from_record_and_relationship(record, relationship, params = {})
-        polymorphic = relationship[:polymorphic]
-
-        return ids_hash(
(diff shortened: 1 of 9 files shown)


TrevorHinesley:
This looks solid to me!

This makes an assumption of note: when an `include` option is specified, if the conditional for the explicitly included relationship evaluates to `false`, then the relationship won't be serialized anyway. In other words, the conditional is the final say, not the specified `include` option. This makes perfect sense to me.

This is also a very helpful feature of AMS that we use in our implementation quite a bit.

Is this good to merged in?

lib/fast_jsonapi/serialization_core.rb
```diff
@@ -82,11 +82,14 @@ def relationships_hash(record, relationships = nil, params = {})
         relationships = relationships_to_serialize if relationships.nil?
 
         relationships.each_with_object({}) do |(_k, relationship), hash|
-          name = relationship[:key]
-          empty_case = relationship[:relationship_type] == :has_many ? [] : nil
-          hash[name] = {
-            data: ids_hash_from_record_and_relationship(record, relationship, params) || empty_case
-          }
+          conditional_proc = relationship[:conditional_proc]
+          if conditional_proc.blank? || conditional_proc.call(record, params)
```
TrevorHinesley: I'm trying to abstract the relationship hash out into it's own class, much like we did in #256, but I'm having a hard time figuring out how to split some of the methods. For instance, `id_hash` would be used inside the `Relationship` class, but is a class method of `SerializationCore`, which is the class that calls the `Relationship`, so it seems a bit like it violates separation of concerns. There are other methods like that too.

@christophersansone since we were discussing the abstraction in the other PR that was merged in, any suggestions for something like that?

christophersansone: @TrevorHinesley Thanks for your help here and for bringing me into the fold.  I'm looking at the relationship code for the first time in a while here, and it has definitely grown more complex.  The idea of extracting `Relationship` just like `Attribute` would be awesome here.

My approach would be to start at `relationships_hash` and make it look just like `attributes_hash`.  It looks like a lot of the methods called within are not called anywhere else, so it should go well.  The only method that appears to be reused elsewhere from my initial read is `id_hash`: it is used by both relationships and the record itself.  However, that method has a few different conditionals, and when the record calls it, it passes in that third parameter to change the behavior.  If we cleanly separated concerns here, I would just give `Relationship` its own `id_hash` method that did exactly what it needed, and just leave the existing `id_hash` alone for now for everyone else to use.

Eventually it looks like if we succeed here, we might eventually wind up with classes like these for `Record`, `Link`, `Include`, etc.  It would definitely make the code easier to reason about and should make most features easier to implement.

shishirmk:
@kyreeves @TrevorHinesley thanks for all the work on this feature 💯

christophersansone:
@kyreeves @TrevorHinesley @shishirmk I just wrapped up the review here as well, and it is awesome work indeed!  Great separation of concerns, a well defined class instead of a hash, and super performant.  👍

kyreeves:
@christophersansone @shishirmk thanks guys!

Issue #250, opened 2018-06-18
&: shortcut doesn't work with custom attributes


kushniryb:
Hey guys 🖖 

First of all - thank you for this awesome gem, great work 👍 
Secondly - I think I may have found a small issue related to the following line - 
https://github.com/Netflix/fast_jsonapi/blob/master/lib/fast_jsonapi/serialization_core.rb#L78

Custom attributes can be defined using following syntax
```ruby
attribute :example do |object|
  object.example?
end
```
Theoretically I should be able to write it as `attribute :example, &:example?` but it doesn't work since `:example?.to_proc.arity #=> -1`

Please let me know what you think about it 🙂

shishirmk:
@kushniryb could you add a test case for this so that we have a good way to look at why it would be failing?

ababich:
I can add here another issue:
`attribute :example, &:computed_example`

where `computed_example` has non-required params like this
```ruby
def computed_example(locale = I18n.default_locale)
  ...
end
```

as a result we will always get issues with `&:` definition

I believe it will work better if by default we will not send `serialization_params` to these methods or maybe introduce condition param for `attribute` to clarify this behavior?

Erol:
@shishirmk Encountered this as well and opened a PR

shishirmk:
This feature is released in [version 1.3](https://github.com/Netflix/fast_jsonapi/releases/tag/1.3)

PR #269, opened 2018-07-13
Allow passing procs with variable arguments when declaring an attribute

Erol:
Allows procs with variable arguments to be used in an attribute declaration, which will also allow the usage of `&:proc`. Instead of having to write something like:

```
class MovieSerializer
  attribute :released_in_year do |object|
    object.release_year
  end
end
```

We can instead write:

```
class MovieSerializer
  attribute :released_in_year, &:release_year
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,18 @@ class MovieSerializer
 end
 ```
 
+Attributes can also use a different name by passing the original method or accessor with a proc shortcut:
+
+```ruby
+class MovieSerializer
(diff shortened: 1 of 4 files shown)


Erol:
Should also address https://github.com/Netflix/fast_jsonapi/issues/250

shishirmk:
@Erol this would be a nice thing to add to 1.3.1 so when this is done i will release that branch

Erol:
@shishirmk Done rebasing from `dev`

Issue #270, opened 2018-07-13
record_type not working for associations


aphilippartd:
Hi,

I am currently facing an issue when trying to include a certain nested record in one of my serialization. I have two serializers: a ContractSerializer and a LegalEntitySerialiser. Each of my legal entities can either be a owner or a tenant. On my Contract model, I have a `has_many :tenants` relationship. My ContractSerializer is defined as follows:

```ruby
class ContractSerializer
  include FastJsonapi::ObjectSerializer

  has_many :tenants, serializer: ::LegalEntitySerializer, record_type: 'tenant'

  attribute :uuid
  attribute :start_date
  attribute :payment_day
  attribute :rent
end
``` 

As shown in the Serializer definition, I would like to set the record_type for tenants to 'tenant' and not 'legal_entity'. However, when trying to run `ContractSerializer.new(Contract.limit(1), {include: ['tenants']}).serializable_hash` i get the following results:

![image](https://user-images.githubusercontent.com/24428618/42693298-007da0ee-86af-11e8-9347-db2315d08aa4.png)

As you can see, the record_type: 'tenant' works on 'relationships' array but not on the 'included' array. I am wondering if this is a normal behaviour of the gem?

Thanks in advance for your help!

TrevorHinesley:
From looking at your code, the record_type should be `legal_entity`. If your tenants are `legal_entity` types, then the record_type should be that, not `tenant`.

shishirmk:
@aphilippartd if you want to change the type in the included record as well, then go ahead and use `set_type` in the LegalEntitySerializer definition.

https://github.com/Netflix/fast_jsonapi#serializer-definition

Erol:
I think this might be related to what I encountered on #271 where inherited serializers do not have a set `record_type` by default.

This is the commit with test case: https://github.com/Netflix/fast_jsonapi/pull/271/commits/e7dc8b4b31857c39295bdd5310e00d6a490ba6ce

aphilippartd:
Thanks for your replies! @Erol , you're right, that's most likely the issue thanks!

PR #271, opened 2018-07-16
Allow the serializer to return sparse fieldsets

Erol:
I also need sparse fieldset support from Fast JSON API. After reading through and seeing that there's no recent activity on #57, #63, and #254, I'd like to take a stab at this.

A client should be able to request only a subset of attributes and relationships using the `fields` option. Here's an example using the movie context from the test suite:

```ruby
options = {
  fields: {
    movie: %i[name actors advertising_campaign],
    actor: %i[name agency]
  },
  include: %i[actors advertising_campaign]
}

```

Note that this would respect both attributes and relationships when the fields have been explicitly given. The resulting `movie` would only return `name` in its `attributes`, and only `actors`, `advertising_campaign` would be retured in its `relationships`.

Fields will also be filtered on included documents. All `actor` will only return `name` and `agency` on their `attributes` and `relationships` respectively.

Records which do not have explicit fields will return all its attributes and relationships, as shown by `advertising_campaign`.

Here's the full JSON:

```json
{
  "data":{
    "id":"232",
    "type":"movie",
    "attributes":{
      "name":"test movie"
    },
    "relationships":{
      "actors":{
        "data":[
          {
            "id":"1",
            "type":"actor"
          },
          {
            "id":"2",
            "type":"actor"
          },
          {
            "id":"3",
            "type":"actor"
          }
        ]
      },
      "advertising_campaign":{
        "data":{
          "id":"1",
          "type":"advertising_campaign"
        }
      }
    }
  },
  "included":[
    {
      "id":"1",
      "type":"actor",
      "attributes":{
        "name":"Test 1"
      },
      "relationships":{
        "agency":{
          "data":{
            "id":"0",
            "type":"agency"
          }
        }
      }
    },
    {
      "id":"2",
      "type":"actor",
      "attributes":{
        "name":"Test 2"
      },
      "relationships":{
        "agency":{
          "data":{
            "id":"1",
            "type":"agency"
          }
        }
      }
    },
    {
      "id":"3",
      "type":"actor",
      "attributes":{
        "name":"Test 3"
      },
      "relationships":{
        "agency":{
          "data":{
            "id":"2",
            "type":"agency"
          }
        }
      }
    },
    {
      "id":"1",
      "type":"advertising_campaign",
      "attributes":{
        "id":1,
        "name":"Movie test movie is incredible!!"
      },
      "relationships":{
        "movie":{
          "data":{
            "id":"232",
            "type":"movie"
          }
        }
      }
    }
  ]
}
```

### DIFF
--- a/lib/fast_jsonapi/fieldset.rb
+++ b/lib/fast_jsonapi/fieldset.rb
@@ -1,7 +1,0 @@
-module FastJsonapi
-  class Fieldset
-    def initialize(fields)
-      @fields = fields
-    end
-  end
-end
--- a/lib/fast_jsonapi/fieldset.rb
+++ b/lib/fast_jsonapi/fieldset.rb
@@ -1,0 +1,7 @@
+module FastJsonapi
+  class Fieldset
+    def initialize(fields)
+      @fields = fields
+    end
+  end
+end
--- a/lib/fast_jsonapi/object_serializer.rb
+++ b/lib/fast_jsonapi/object_serializer.rb
@@ -41,7 +41,7 @@ module FastJsonapi
 
       return serializable_hash unless @resource
 
-      serializable_hash[:data] = self.class.record_hash(@resource, @fieldsets[self.class.reflected_record_type.to_sym], @params)
+      serializable_hash[:data] = self.class.record_hash(@resource, @fieldsets[self.class.record_type.to_sym], @params)
       serializable_hash[:included] = self.class.get_included_records(@resource, @includes, @known_included_objects, @fieldsets, @params) if @includes.present?
       serializable_hash
     end
@@ -51,7 +51,7 @@ module FastJsonapi
 
       data = []
       included = []
-      fieldset = @fieldsets[self.class.reflected_record_type.to_sym]
+      fieldset = @fieldsets[self.class.record_type.to_sym]
       @resource.each do |record|
         data << self.class.record_hash(record, fieldset, @params)
         included.concat self.class.get_included_records(record, @includes, @known_included_objects, @fieldsets, @params) if @includes.present?
--- a/lib/fast_jsonapi/object_serializer.rb
+++ b/lib/fast_jsonapi/object_serializer.rb
@@ -119,6 +119,7 @@ module FastJsonapi
         subclass.race_condition_ttl = race_condition_ttl
         subclass.data_links = data_links
         subclass.cached = cached
+        subclass.set_type(subclass.reflected_record_type) if subclass.reflected_record_type
       end
 
       def reflected_record_type
--- a/lib/fast_jsonapi/object_serializer.rb
+++ b/lib/fast_jsonapi/object_serializer.rb
@@ -7,6 +7,7 @@ require 'fast_jsonapi/attribute'
 require 'fast_jsonapi/relationship'
 require 'fast_jsonapi/link'
 require 'fast_jsonapi/serialization_core'
+require 'fast_jsonapi/fieldset'
 
 module FastJsonapi
   module ObjectSerializer
@@ -41,8 +42,8 @@ module FastJsonapi
 
       return serializable_hash unless @resource
 
-      serializable_hash[:data] = self.class.record_hash(@resource, @params)
-      serializable_hash[:included] = self.class.get_included_records(@resource, @includes, @known_included_objects, @params) if @includes.present?
+      serializable_hash[:data] = self.class.record_hash(@resource, @fieldsets[self.class.reflected_record_type.to_sym], @params)
+      serializable_hash[:included] = self.class.get_included_records(@resource, @includes, @known_included_objects, @fieldsets, @params) if @includes.present?
       serializable_hash
     end
 
@@ -51,9 +52,10 @@ module FastJsonapi
 
       data = []
       included = []
+      fieldset = @fieldsets[self.class.reflected_record_type.to_sym]
       @resource.each do |record|
-        data << self.class.record_hash(record, @params)
-        included.concat self.class.get_included_records(record, @includes, @known_included_objects, @params) if @includes.present?
+        data << self.class.record_hash(record, fieldset, @params)
+        included.concat self.class.get_included_records(record, @includes, @known_included_objects, @fieldsets, @params) if @includes.present?
       end
 
       serializable_hash[:data] = data
@@ -70,6 +72,8 @@ module FastJsonapi
     private
 
     def process_options(options)
+      @fieldsets = deep_symbolize(options[:fields].presence || {})
+
       return if options.blank?
 
       @known_included_objects = {}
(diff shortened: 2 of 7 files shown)


Erol:
Hope you can have a look at this as well @shishirmk

shishirmk:
will take a look at this and get back to you in a day..

Erol:
Thanks. Let me know if there's anything else I need to consider

lib/fast_jsonapi/fieldset.rb
```diff
@@ -0,0 +1,7 @@
+module FastJsonapi
+  class Fieldset
+    def initialize(fields)
+      @fields = fields
```
shishirmk: could we run define and run deep_symbolize method here? I think it makes sense for it to be transparent to the rest of the code.

lib/fast_jsonapi/object_serializer.rb
```diff
@@ -41,8 +42,8 @@ def hash_for_one_record
 
       return serializable_hash unless @resource
 
-      serializable_hash[:data] = self.class.record_hash(@resource, @params)
-      serializable_hash[:included] = self.class.get_included_records(@resource, @includes, @known_included_objects, @params) if @includes.present?
+      serializable_hash[:data] = self.class.record_hash(@resource, @fieldsets[self.class.reflected_record_type.to_sym], @params)
```
shishirmk: `reflected_record_type` is the meaningful default it can be overridden in `set_type` so it probably makes sense here to use `self.record_type`

lib/fast_jsonapi/object_serializer.rb
```diff
@@ -51,9 +52,10 @@ def hash_for_collection
 
       data = []
       included = []
+      fieldset = @fieldsets[self.class.reflected_record_type.to_sym]
```
shishirmk: `reflected_record_type` is the meaningful default it can be overridden in `set_type` so it probably makes sense here to use `self.record_type`

lib/fast_jsonapi/fieldset.rb
Erol: That's a remnant of a previous attempt and no longer needed. I'll delete it and remove the reference

lib/fast_jsonapi/object_serializer.rb
Erol: 👍

lib/fast_jsonapi/object_serializer.rb
Erol: 👍

lib/fast_jsonapi/object_serializer.rb
```diff
@@ -104,6 +120,7 @@ def inherited(subclass)
         subclass.race_condition_ttl = race_condition_ttl
         subclass.data_links = data_links
         subclass.cached = cached
+        subclass.fieldset = fieldset.dup if fieldset.present?
```
shishirmk: fieldsets needs to be passed in each time `serializable_hash` is invoked. I don't think it makes sense to inherit this when subclassing.

lib/fast_jsonapi/serialization_core.rb
```diff
@@ -40,36 +41,39 @@ def links_hash(record, params = {})
         end
       end
 
-      def attributes_hash(record, params = {})
-        attributes_to_serialize.each_with_object({}) do |(_k, attribute), hash|
+      def attributes_hash(record, fieldset = nil, params = {})
+        attributes = attributes_to_serialize
+        attributes = attributes.slice(*fieldset) if fieldset.present?
```
shishirmk: running this slice for each record in an array is probably unnecessary. It would be nice if we can avoid that. Not needed for merging in.

lib/fast_jsonapi/serialization_core.rb
```diff
@@ -40,36 +41,39 @@ def links_hash(record, params = {})
...
+        attributes.each_with_object({}) do |(_k, attribute), hash|
           attribute.serialize(record, params, hash)
         end
       end
 
-      def relationships_hash(record, relationships = nil, params = {})
+      def relationships_hash(record, relationships = nil, fieldset = nil, params = {})
         relationships = relationships_to_serialize if relationships.nil?
+        relationships = relationships.slice(*fieldset) if fieldset.present?
```
shishirmk: running this slice for each record in an array is probably unnecessary. It would be nice if we can avoid that. Not needed for merging in.

shishirmk:
@Erol This looks really good. Thank you for the work you have put in. I have added a few comments please address them. Add how to use this feature to the read me. Then we can merge it in and I will release it in the next version by the end of the week.

lib/fast_jsonapi/object_serializer.rb
Erol: @shishirmk Seems like `record_type` is not being set for inherited serializers. Got this when I introduced a spec for it:

```
  1) FastJsonapi::ObjectSerializer sets the correct record type
     Failure/Error: expect(EmployeeSerializer.record_type).to eq 'employee'

       expected: "employee"
            got: nil

       (compared using eql?)
     # ./spec/lib/object_serializer_inheritance_spec.rb:99:in `block (2 levels) in <top (required)>'
```

I'll write a fix for that as well.

lib/fast_jsonapi/object_serializer.rb
Erol: That's another remnant I failed to cleanup. Should be gone after the next commit.

lib/fast_jsonapi/object_serializer.rb
Erol: @shishirmk Looks like `record_type` is not being set for inherited serializers. I added a spec for it and got this:

```
  1) FastJsonapi::ObjectSerializer sets the correct record type
     Failure/Error: expect(EmployeeSerializer.record_type).to eq 'employee'

       expected: "employee"
            got: nil

       (compared using ==)
     # ./spec/lib/object_serializer_inheritance_spec.rb:99:in `block (2 levels) in <top (required)>'
```

Will write a fix for that as well.

shishirmk: Yup go ahead and fix that as well.

lib/fast_jsonapi/serialization_core.rb
Erol: Yeah that's the ideal. I'll open another PR as an improvement.

Erol:
@shishirmk Addressed your feedback with the commits above. I've also added instructions on the README.

Also added a fix to set the `record_type` on inherited serializers here: https://github.com/Netflix/fast_jsonapi/commit/e7dc8b4b31857c39295bdd5310e00d6a490ba6ce. That might partially resolve #270.

g13ydson:
@Erol @shishirmk I have a problem. I have created the following serializers:
```ruby
class UserSerializer
  attributes :name_badge: 
  has_many :occupations
end
```

```ruby
class OccupationSerializer
  attributes :type
end
```

When I try:
`UserSerializer.new(@user, {include: [:occupations], fields: {user: [:name]}})
`

I get as return the JSON below, notice that the relationships field is empty, should not come filled?

```json
{
    "data": {
        "id": "1",
        "type": "user",
        "attributes": {
            "name": "Gleydson"
        },
        "relationships": {}
    },
    "included": [
        {
            "id": "1",
            "type": "occupation",
            "attributes": {
                "type": "Teacher"
            }
        }
    ]
}
```

shishirmk:
I dont think relationships should be filled in according to the spec. Here is the link http://jsonapi.org/format/#fetching-sparse-fieldsets
> If a client requests a restricted set of fields for a given resource type, an endpoint MUST NOT include additional fields in resource objects of that type in its response.

Issue #237, opened 2018-06-06
`set_id` issue with included


ababich:
Picture simple relationship like this:
```
class ASerializer
  include FastJsonapi::ObjectSerializer

  has_many :bs
end

class BSerializer
  include FastJsonapi::ObjectSerializer

  set_id :custom_id
end
```

The problem is that `fast_jsonapi`to serialize A objects with B included will generate structure like this:
```
{
  data: [
    { id: AID,
      type: :a,
      attributes: {},
      relationships: {
       bs: [
         { id: B1ID, type: :b },
         { id: B2ID, type: :b }
       ]       
      }
  ],
  included: [
    { id: B1CUSTOMID, ... },
    { id: B2CUSTOMID, ... },
  ]
}
```

So that unless we specifically set `id_method_name` in relationship and implement it accordingly this makes no sense

I suggest to add additional checks to prevent this situation, i.e. have `included` ids not matching `relationships` ids

----

This is backlog feature request, in general, as it improves API but not changes functionality and may affect performance

shishirmk:
This is interesting. I believe this could be a bug. Do you mind writing a failing test to better capture this issue? I can try to fix it if you could do that.

ababich:
Sure, please take a look at https://github.com/ababich/fast_jsonapi/compare/master...set_id_fail

yale:
hi folks - just bumping this as we're running into this same issue as well.

shishirmk:
@yale there is a workaround for this as @ababich mentioned you need to set both `id_method_name` and `set_id`. We would need to figure out how to `set_id` so that it can be used both in the serializer and all the places the record is used in a relationship

shishirmk:
Released in [Version 1.4](https://github.com/Netflix/fast_jsonapi/releases/tag/1.4)

zinosama:
@shishirmk I don't think this issue is resolved by #274 since I'm still running into the same problem with 1.4. As manojmj92 [pointed out](https://github.com/Netflix/fast_jsonapi/pull/274#issuecomment-406156094), #274 only solves the issue with `id_method_name` not being set with block.

PR #274, opened 2018-07-18
Evaluate ids via the specified 'id_method_name' when relationships are evaluated via a block

manojmj92:
Current behaviour:

Currently, we are able to use the `id_method_name` argument to specify the method which has to be used to evaluate the id of an object.

Eg:

```
class Movie
 has_many :actors, id_method_name: :actor_uuids
end
```
This works as expected.

However, when these relationships are evaluated via block, `id_method_name` is not taken into account while evaluating the `id` of the object (it automatically evaluates `id` from the `id` method of the object, no matter what method we specify in `id_method_name`)

So, like
```
class Movie
 has_many :actors, id_method_name: :uuid do |movie|
  movie.actors
 end
end
```
does not set the `uuid` of the actor as it's `id` even when we have specified `id_method_name` as `uuid`

Change:

This PR is aimed at solving the problem mentioned above.

With this change, when relationships are evaluated via a block and if `id_method_name` is specified, the id will be inferred from `id_method_name`.

If `id_method_name` is not specified, it defaults back to getting `id` from the object's `id` method.


ababich:
Does it address #237 ? Please, see example of rspec for the case https://github.com/ababich/fast_jsonapi/compare/master...set_id_fail

manojmj92:
@ababich Thanks for the message! This PR is not aimed at solving #237. In fact, I also faced the problem mentioned in #237, but got around to solving it by specifying both `id_method_name` and `set_id` to point to the same method.

spec/lib/object_serializer_class_methods_spec.rb
```diff
@@ -87,6 +87,28 @@
...
+    end
+
+    after do
+      MovieSerializer.relationships_to_serialize.delete(:awards)
+    end
+
+    context 'awards is not included' do
+      subject(:hash) { MovieSerializer.new(movie).serializable_hash }
+
+      it 'returns correct hash' do
```
ababich: from the test specification it is not clear that [or why] the hash is actually *correct*

is it possible to compare `hash[:data][:relationships][:awards][:data]` to a full hash where it will be visible that exactly `imdb_award_id` was used to serialize `id`

manojmj92: Sure, I'll work on this!

manojmj92:
@ababich - Thank you for the very thoughtful review! :) I've made the necessary changes. Can you check and see if it's okay?

ababich:
@manojmj92 I have no more questions to it at the moment. Thanks! Anyway, you'll need @shishirmk to go forward

manojmj92:
@ababich Thanks for the review! :)

manojmj92:
@shishirmk I went ahead and refactored the code to use `id_method_name`. It looks much cleaner now, thank you for your feedback! :).
Merge conflicts have also been resolved.

shishirmk:

PR #275, opened 2018-07-18
Implement polymorphic serialization for includes

jshow:
Implement polymorphic serialization for includes

### DIFF
--- a/spec/lib/object_serializer_spec.rb
+++ b/spec/lib/object_serializer_spec.rb
@@ -251,12 +251,6 @@ describe FastJsonapi::ObjectSerializer do
         expect(movies_serialized).to include(movie.id)
       end
     end
-
-    it 'polymorphic throws an error that polymorphic is not supported' do
-      options = {}
-      options[:include] = [:groupees]
-      expect(-> { GroupSerializer.new([group], options)}).to raise_error(NotImplementedError)
-    end
   end
 
   context 'when testing included do block of object serializer' do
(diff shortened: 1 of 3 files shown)


spec/lib/object_serializer_spec.rb
```diff
@@ -251,12 +251,6 @@ def advertising_campaign
         expect(movies_serialized).to include(movie.id)
       end
     end
-
-    it 'polymorphic throws an error that polymorphic is not supported' do
```
lassiter: Are you planning on replacing with tests for polymorphic relationships?

shishirmk:
@jshow could you add some tests please. I can merge it in once tests are added

jshow:
will do @shishirmk @lassiter - was looking for feedback on impl first. cheers

jshow:
Just getting into writing tests, but appears that polymorphic doesn't have any present tests ? 

Doesn't seem possible, but scanning the code it appears to be so

Gonna take me a quite a bit longer to reconstruct the polymorphic spec - can I have some help beyond the docs ?

lassiter:
If I recall, there is a context called `groupee` but there are no unit tests.

shishirmk:
@jshow we went with different pull request. So closing this. Polymorphic includes are now supported and Released in [Version 1.4](https://github.com/Netflix/fast_jsonapi/releases/tag/1.4)

jshow:
thanks folks, I got buried

PR #280, opened 2018-07-20
Introduce the ability to add `meta` tag for every resource in the collection

manojmj92:
Current behaviour:

We currently allow the use of `meta` tags, but this is limited for use only at a top/collection level, that has to be passed in along with the call to `serialized_json`, for eg:

```ruby
options = {}
options[:meta] = { total: 2 }
MovieSerializer.new([movie], options).serialized_json
```

Proposed change:

[JSON API spec](http://jsonapi.org/format/#document-resource-objects) specifies that `meta` tags can also be used at a `resource` level for non-standard information that cannot be included as attributes or relationships for each resource inside the collection

So like,

```
{
  "data": [
   {
     "id": "3",
     "type": "movie",
     "attributes": {
       "name": "Test movie 1",
       "year": 2008
     },
	 "meta": {
	   "years_since_release": 10
	 }
   },
   {
     "id": "4",
     "type": "movie",
     "attributes": {
       "name": "Test movie 2",
       "year": 2007
     },
	 "meta": {
	   "years_since_release": 11
	 }
   }
}
```
is a valid response as per JSON API SPEC. However, we do not allow resource level meta tags right now and this PR is aimed at providing this feature.

With this feature, the above response can be obtained with a serializer like,

```ruby
class MovieSerializer

  attributes :name, :year

  meta do |movie|
	{
	  years_since_release: Date.current.year - movie.year
	}
  end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,23 @@ class MovieSerializer
 end
 ```
 
+### Meta Per Resource
+
+For every resource in the collection, you can include a meta object containing non-standard meta-information about a resource that can not be represented as an attribute or relationship.
+
+
+```ruby
+class MovieSerializer
+  include FastJsonapi::ObjectSerializer
+
+  meta do |movie|
+    {
+      years_since_release: Date.current.year - movie.year
+    }
+  end
+end
+```
+
 ### Compound Document
 
 Support for top-level and nested included associations through ` options[:include] `.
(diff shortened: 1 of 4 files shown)


shishirmk:
@manojmj92 Please fix the conflicts. It looks good for the most part. I can merge it in once the conflicts are fixed

manojmj92:
@shishirmk Done! :)

Erol:
I opened a discussion on #292 to explore if we can change the syntax a bit. Would you be able to take a look @manojmj92 @shishirmk?

shishirmk:

PR #283, opened 2018-07-22
Fix confusing error when using include and no relationship was set

aledalgrande:
If you forgot to set any `has_many` in the serializer and tried to serialize with an `include` you would get:
```
NoMethodError (undefined method `[]' for nil:NilClass):
```

(no backtrace from the gem) That is not very helpful.

Setting the variable with a default case makes sure the right error message gets displayed.


aledalgrande:
Is this project still maintained?

shishirmk:
@aledalgrande it is maintained. I will take a look at your fix this week and give you a reply.

shishirmk:
@aledalgrande thanks for the fix.

shishirmk:

PR #295, opened 2018-08-08
Improve key search for camelized keys

zinovyev:
Allow to define a `model_name` and `fieldsets` with their underscore names even if the `set_key_transform` is set to `:camel`.

**Should be valid**:

```ruby
MovieSerializer.set_key_transform :camel
serializer = MovieSerializer.new(movie, { fields: { movie_model_name: [:attribute_name] } })
```

**As well as**:

```ruby
MovieSerializer.set_key_transform :camel
serializer = MovieSerializer.new(movie, { fields: { MovieModelName: [:AttributeName] } })
```

### DIFF
--- a/lib/fast_jsonapi/serialization_core.rb
+++ b/lib/fast_jsonapi/serialization_core.rb
@@ -41,11 +41,16 @@ module FastJsonapi
       end
 
       def attributes_hash(record, fieldset = nil, params = {})
-        attributes = attributes_to_serialize
-        attributes = attributes.slice(*fieldset) if fieldset.present?
+        attributes = slice_attributes(attributes_to_serialize, fieldset)
         attributes.each_with_object({}) do |(_k, attribute), hash|
           attribute.serialize(record, params, hash)
         end
+      end
+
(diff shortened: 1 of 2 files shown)


lib/fast_jsonapi/serialization_core.rb
```diff
@@ -41,13 +41,18 @@ def links_hash(record, params = {})
...
-        attributes = attributes.slice(*fieldset) if fieldset.present?
+        attributes = slice_attributes(attributes_to_serialize, fieldset)
         attributes.each_with_object({}) do |(_k, attribute), hash|
           attribute.serialize(record, params, hash)
         end
       end
 
+      def slice_attributes(attributes, fieldset)
+        return attributes unless fieldset.present?
+        fieldset = fieldset.map { |key| run_key_transform(key) }
```
shishirmk: This might not be performant to run on serialization of each object. I wonder if we can do all of this once per instantiation of the serializer.

shishirmk:
We dont want to support this feature as we think this can be handled in the controller.

PR #299, opened 2018-08-10
add the option "key" in customizable options in the README

g13ydson:
There is an option called "key" that does not exist in the documentation. I had to look in the code to know that it exists. With this option, you can change the json object key.

`belongs_to: owner, key: :user `

```
{
  "data": {
    "id": "3",
    "type": "movie",
    "attributes": {
      "name": "test movie",
      "year": null
    },
    "relationships": {
      "user": {
        "data": {
          "id": "3",
          "type": "owner"
        }
      }
    }
  }
}
```


shishirmk:

Issue #303, opened 2018-08-17
Flipped ArgumentError with conditional attributes


coconup:
Conditional attributes are throwing me an `ArgumentError` if I don't give two arguments to the `Proc` passed to the `if:` keyword:

```ruby
  attribute :my_attribute, if: ->(obj) { obj.my_condition }
```

returns `ArgumentError: wrong number of arguments (given 2, expected 1)` (shouldn't it be the other way around, if anything?)

Passing a second argument, on the other hand, makes things work just fine:

```ruby
  attribute :my_attribute, if: ->(obj, _) { obj.my_condition }
```

shishirmk:
@coconup  if params are not passed in we should probably set it to a default empty hash. Is this something you can make a pull request for ?

coconup:
@shishirmk here you go: https://github.com/Netflix/fast_jsonapi/pull/305

PR #305, opened 2018-08-23
Optional attributes lambda support with 1 or 2 arguments

coconup:
Fixes [issue 303](https://github.com/Netflix/fast_jsonapi/issues/303)

### DIFF
--- a/lib/fast_jsonapi/attribute.rb
+++ b/lib/fast_jsonapi/attribute.rb
@@ -20,7 +20,7 @@ module FastJsonapi
 
     def include_attribute?(record, serialization_params)
       if conditional_proc.present?
-        conditional_proc.call(record, serialization_params)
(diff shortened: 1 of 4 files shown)


lib/fast_jsonapi/attribute.rb
```diff
@@ -20,7 +20,7 @@ def serialize(record, serialization_params, output_hash)
 
     def include_attribute?(record, serialization_params)
       if conditional_proc.present?
-        conditional_proc.call(record, serialization_params)
+        conditional_proc.arity.abs == 1 ? conditional_proc.call(record) : conditional_proc.call(record, serialization_params)
```
shishirmk: Since this check would run when serializing every object it could be expensive. Instead could we explore just setting a default `serialization_params = {}`. That might be faster and might solve the problem.

shishirmk:
We dont want to support this feature as we think this can be handled in the definition of the proc.

PR #312, opened 2018-09-03
Do not use ActiveSupport core extensions

rintaun:
Core extensions often don't play well with many other gems... considering that they only seem to be included for one `to_json` call, it would probably be better to avoid using them altogether and call `ActiveSupport::JSON.encode` directly.


shishirmk:

PR #313, opened 2018-09-09
Fix set_key_transform's set_type to give priority to pre-set value

### LINKED ISSUE #310: Key transform resets type in fast_jsonapi 1.3
In fast_jsonapi 1.3 (but not in 1.2), calling `set_key_transform` after `set_type` seems to reset the type

```
class DeckSerializer
  set_type :decks
  set_key_transform :dash
end

DeckSerializer.new(Deck.first).serializable_hash
=> {:data=>{:id=>1, :type=>:deck}}
```

Note the singular "deck". I expected the plural "decks".

If I switch the order of `set_type` and `set_key_transform`, it works.

sakuraineed:
fixes #310

### DIFF
--- a/lib/fast_jsonapi/object_serializer.rb
+++ b/lib/fast_jsonapi/object_serializer.rb
@@ -143,7 +143,11 @@ module FastJsonapi
         self.transform_method = mapping[transform_name.to_sym]
 
         # ensure that the record type is correctly transformed
-        set_type(reflected_record_type) if reflected_record_type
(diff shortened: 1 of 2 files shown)


shishirmk:
#310 is about singular and plural forms of the record_type. Could you add a spec for that too?

sakuraineed:
@shishirmk Thanks for the comment.
I added spec for singular and plural, so I removed same checking example.

PR #326, opened 2018-09-26
added spec for has_one-through relationship

gorenje:
I'm debugging an issue with [rabl](https://rubygems.org/gems/rabl) in combination with fast_jsonapi within a (private) Rails project. 

It occurs when requiring the fast_jsonapi gem and has something to do with a has_one through relationship in the model being serialized, or with rabl and how it handles attributes, or something completely different...

This spec is just a check that fast_jsonapi can also handle has_one through relationships, and it can!


shishirmk:

PR #329, opened 2018-10-02
Improve documentation with id_method_name

zinosama:
```ruby
# invokes b_id_method on the resource of ASerializer, presumably one or more instances of A
ASerializer.has_many(:bs, id_method_name: :b_id_method) 

# invokes b_id_method on each instance of `a.bs`, presumably instances of B
ASerializer.has_many(:bs, id_method_name: :b_id_method) do |a|
  a.bs
end
```
This behavior was quite subtle to me and took me some investigation to understand. I don't know if this is by design but I think it should at least be documented.


shishirmk:
Thanks for working on this

PR #330, opened 2018-10-03
Don't share data_links among inherited serializers.

matzko:
As demonstrated by the included specs, if serializers share a common ancestor that defines data_links, they end up sharing the same data_links object, overwriting each other, so that the last-evaluated sub-class overwrites the others' data_links values.


shishirmk:

PR #331, opened 2018-10-05
Allow block for id customization

### LINKED ISSUE #315: Allow blocks for setting id
Currently we allow to customize any attribute by using blocks. I would like to extend the same behaviour to ids. Currently the only way to customize id is by using `set_id :custom_id` but this specifies an alternative method in the object. Instead, I would like to be able to customize id right here in the serializer, using something like this:
```
class MovieSerializer

  attributes :name, :year
  set_id do |record|
    "This will be a string ID with #{record.id} inside"
  end
```
This is useful in situations when you have different types of objects (say, `Movie`, `MovieBuilder`, `BaseMovie`) being serialized using the same serializer (`GeneralMovieSerializer`), and you want to unify their ids setting code. What do you think?

larissa:
Allow an ID of object to be customized directly on the serializer by
passing a block to `set_id` as opposed to only through a model property.

We already allow for attributes that do not have a model property of the
same name to be customized directly on the serializer using a block.

Closes #315

### DIFF
--- a/README.md
+++ b/README.md
@@ -427,7 +427,7 @@ Option | Purpose | Example
 ------------ | ------------- | -------------
 set_type | Type name of Object | ```set_type :movie ```
 key | Key of Object | ```belongs_to :owner, key: :user ```
-set_id | ID of Object | ```set_id :owner_id ```
+set_id | ID of Object | ```set_id :owner_id ``` or ```set_id { |record| "#{record.name.downcase}-#{record.id}" }```
 cache_options | Hash to enable caching and set cache length | ```cache_options enabled: true, cache_length: 12.hours, race_condition_ttl: 10.seconds```
 id_method_name | Set custom method name to get ID of an object | ```has_many :locations, id_method_name: :place_ids ```
 object_method_name | Set custom method name to get related objects | ```has_many :locations, object_method_name: :places ```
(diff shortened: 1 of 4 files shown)


manojmj92:
@larissa thanks for the PR! Can you please add examples in the PR description on how this change will be used in a serialzier? It'd be very helpful.

larissa:
@manojmj92 I've just updated the commit message, hope that helps!

manojmj92:
@larissa Thanks for the update! By the looks of it, this is a very useful change. I think this should go in the documentation as well. What do you think? Can you update the docs as well? I'd be happy to do it if you want. :)

larissa:
@manojmj92 I wasn't sure where to put it in the docs since there isn't a section for `set_id`, so I just added as a example in the customizable options table in the README. I can add a full section for `set_id` if you'd like too.

Also is there any other documentation you were referring to or was it just the README file?

manojmj92:
@larissa thanks for the update! This looks good!

wilsonsilva:
What needs to happen for this to be merged?

shishirmk:

JohnBDonner:
@shishirmk @larissa Thanks for merging this PR. I noticed this still doesn't exactly follow the same behaviour as with `attributes`. You can't pass a custom `params` to the block like you can with `attributes`. It seems the code for handling `set_id` is quite separate from `attributes` but would it be possible to achieve the following?

```ruby
class SomeSerializer

  set_id :id do |_object, params| # :id may be unnecessary
    params[:id]
  end
end
```

PR #334, opened 2018-10-09
require AS core extension for time

hmcfletch:
`5.minutes` was failing when i ran the performance spec


shishirmk:

PR #339, opened 2018-10-18
Fix params not being hash by default

mhluska:
This fixes a minor thing that has been nagging at me for a while: I always had to do `params && params[:foo]` in an attribute block because params is undefined unless the controller explicitly passes it. I think we should be able to expect params to always be defined.

The bug occurs when no options are passed and the early return `return if options.blank?` prevents params being set to `{}` by default.


manojmj92:
Fantastic! Thank you for this fix! :)

PR #355, opened 2018-11-26
Fix markdown rendering on customizable options table set id row

kuntoaji:
before:
before

after:
after


shishirmk:
Thanks but already fixed in dev

PR #362, opened 2018-12-19
Preset Sparse Fieldsets

Bhacaz:
I was thinking about a new feature base on the _Sparse Fieldsets_ option.

It consist to declared some presets of fields in the _Serializer_ class so it can be easily reuse, instead of passing an option hash or declaring multiple serializer for each case (i.e. `AccountSerializer`, `MinimalAccountSerializer`)

There is an example of use case.

```ruby
class AccountSerializer

  attributes :first_name,
             :last_name,
             :birthday,
             :email,
             :picture_url

  field_preset :minimal, :first_name, :last_name
  field_preset :profile, :first_name, :last_name, :birthday, :picture_url
end
```

```ruby
# GET /me/friends
AccountSerializer.new(auth_account, { fields: { account: [:first_name, :last_name] } }) # using Sparse Fieldsets
# OR
AccountSerializer.new auth_account.friends, field_preset: :minimal # using Field preset

# GET /me/profile
AccountSerializer.new(auth_account, { fields: { account: [ :first_name, :last_name, :birthday, :picture_url] } }) # using Sparse Fieldsets
# OR
AccountSerializer.new auth_account.friends, field_preset: :profile # using Field preset
```

### DIFF
--- a/lib/fast_jsonapi/object_serializer.rb
+++ b/lib/fast_jsonapi/object_serializer.rb
@@ -89,6 +89,8 @@ module FastJsonapi
         @includes = options[:include].delete_if(&:blank?).map(&:to_sym)
         self.class.validate_includes!(@includes)
       end
+
+      @fieldsets[self.class.record_type.to_sym] = self.class.fields_presets[options[:preset_fields]] if options[:preset_fields]
     end
 
     def deep_symbolize(collection)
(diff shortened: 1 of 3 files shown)


Bhacaz:
@y0nigt It's an excellent question. But it'll have the same behavior when using [Sparse fieldsets](https://github.com/Netflix/fast_jsonapi#sparse-fieldsets). 
Maybe it can be report as an issue?

shishirmk:
We dont want to support this feature as we think this can be handled in the controller.

Issue #365, opened 2018-12-21
can't modify frozen Array


dapi:
Error raised when we pass frozen array to `include` option.

https://github.com/Netflix/fast_jsonapi/blob/fdcaed6f0d72cf793b3117a5081de5fbd7466574/lib/fast_jsonapi/object_serializer.rb#L89

PR #366, opened 2018-12-21
[#365] Support frozen array in  option

dapi:
Fix error raised when we pass frozen array to `include` option.

https://github.com/Netflix/fast_jsonapi/blob/fdcaed6f0d72cf793b3117a5081de5fbd7466574/lib/fast_jsonapi/object_serializer.rb#L89


shishirmk:
Thanks for the PR 👍

Issue #301, opened 2018-08-15
Support automatic pluralization of resource types


amysutedja:
AMS produces pluralized resource types by default for all serialized types, whereas fast_jsonapi does not. This makes migration from AMS to fast_jsonapi cumbersome, especially if you have a lot of types.

My original plan was to just write some code to perform the pluralization transform automatically, but it turns out that `FastJsonapi::ObjectSerializer` calls `set_type` rather aggressively, which results in my automatic pluralization changes being either ignored or clobbered.

What do you all think about just supporting pluralization of the type directly?

shishirmk:
@amysutedja why don't you add a new option to the serializer called `pluralize_type: true` by default let it be false that way existing behavior would be retained. Is this something you could contribute?

manojmj92:
@shishirmk @amysutedja I'd be happy to take this!

shishirmk:
@manojmj92 go ahead!

brandonjyuhas:
@manojmj92 hey! Thanks for taking time to work on this enhancement. Any updates on the automatic pluralization feature (or any idea of a workaround)?

Right now my team and I are setting each relationship's `record_type` manually, and we'd love to do it in a more programmatic way.

Thanks again!

amysutedja:
@manojmj92 If you haven’t taken a look at this yet, I have some ideas on a potential fix.

manojmj92:
I've not taken this up. Please feel free to work on this.

ritesh2741:
@amysutedja  can you please share your idea on this or if there is any update

PhilT:
I needed this as well so here's what I came up with:

```
module FastJsonapi
  module ObjectSerializer
    class_methods do
      def pluralize(type_name)
        type_name.to_s.pluralize
      end

      alias_method :original_set_type, :set_type
      def set_type(type_name) # rubocop:disable Naming/AccessorMethodName
        original_set_type pluralize(type_name)
      end

      alias_method :original_belongs_to, :belongs_to
      def belongs_to(type_name, options = {})
        options[:record_type] = pluralize(type_name)
        original_belongs_to type_name, options
      end

      alias_method :original_has_many, :has_many
      def has_many(type_name, options = {}) # rubocop:disable Naming/PredicateName
        options[:record_type] = pluralize(type_name)
        original_has_many type_name, options
      end
    end
  end
end
```

glyoko:
[...]

PR #376, opened 2019-01-12
Support automatic pluralization of types

amysutedja:
Resolves https://github.com/Netflix/fast_jsonapi/issues/301

Enables support for `pluralize_type` in serializer definitions and relationship definitions, which pluralizes the type as expected.

Additional Note: In theory it's possible that we could try to infer whether relationship types should be pluralized by asking the relationship record's serializer about its pluralization state. But currently the code only interacts with the relationship record very lightly, and I was concerned about potentially missing performance targets for complicated records, so I didn't research this path very extensively.


spec/lib/object_serializer_class_methods_spec.rb
```diff
@@ -486,4 +486,63 @@ def year_since_release_calculator(release_year)
...
+    context 'when pluralize is false' do
+      let(:pluralize) { false }
+
+      it 'returns correct hash which type equals non-pluralized value' do
+        expect(serializable_hash[:data][:type]).to eq :movie
+      end
+    end
+  end
+
+  describe '#pluralize_type after #set_type' do
```
glyoko: As it stands there's an issue where type is not correctly pluralized in relationships. The following test shows this:

```
    context 'when pluralizing a relationship type after #set_type' do
      before do
        ActorSerializer.pluralize_type true
      end

      after do
        ActorSerializer.pluralize_type nil
      end

      it 'returns correct hash which relationship type equals transformed set_type value' do
        expect(serializable_hash[:data][:relationships][:actors][:data][0][:type]).to eq(:actors)
        expect(serializable_hash[:included][:type]).to eq(:actors)
      end
    end
```

This also means there's a mismatch between relationship and include types. Currently:

```
> serializable_hash[:data][:relationships][:actors][:data][0][:type]
# => :actor
> serializable_hash[:included][:type]
# => :actors
```

amysutedja: Unfortunately, as far as I understand, `fast_jsonapi` does not actually "know" internally what serializer is associated with any given relationship. The relationship data is built via simply reflecting on the object and constructing the abbreviated relationship structure. In theory it's possible to make it aware of the serializer, but see my original "Additional Note" -- I ended up being worried about missing performance targets, so I didn't implement a change here.

In order to make this scenario work, I implemented the ability to mark a specific relationship as plural with `pluralize_type: true`. (Check out `object_serializer_pluralization_spec.rb`.) This isn't ideal, but it solves this scenario in lieu of a higher-level decision about perf targets.

spec/lib/object_serializer_class_methods_spec.rb
```diff
@@ -486,4 +486,63 @@ def year_since_release_calculator(release_year)
...
+      MovieSerializer.pluralize_type true
+    end
+
+    after do
+      MovieSerializer.pluralize_type nil
+      MovieSerializer.set_type :movie
+    end
+
+    context 'when sets singular type name' do
+      let(:type_name) { :film }
```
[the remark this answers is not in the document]
amysutedja: Can you explain more about why this should be the case?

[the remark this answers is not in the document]
glyoko: Look at line 541. That's directly setting the type to be plural in the before block, before the `MovieSerializer.pluralize_type true` line is even hit. In fact, you could comment out line 524 and this test would still pass.

Given that, both examples should be given a singular type name for these tests to check anything useful. In fact, you could even get rid of the let altogether and just say `MovieSerializer.set_type :film` on line 523.

spec/lib/object_serializer_class_methods_spec.rb
glyoko: I sort of glossed over that note when I first reviewed this PR, and after banging my head against this for a bit, I'm coming to the same conclusion... I'm going to investigate this a bit more because it is important for my use case, but I understand that the complexity involved in such a refactor may not be worth the effort.

At any rate it may not be worth including in this PR, as it already achieves the majority of what it sets out to do.

sdhull:
This would help us move our serializers over to fast_jsonapi. Any update on this PR?

amysutedja:
@sdhull I haven't heard back on any suggested changes from the maintainers. And the Travis builds break, for some reason.

You might want to note your interest in the original ticket, #301.

zdavis:
I wonder if you'd be willing to open this PR against https://github.com/fast-jsonapi/fast_jsonapi, which seems to me to be the way forward with this library, since the maintainers aren't responsive here. The travis failures seem completely unrelated to your changes, and I'd love to see this PR merged into that fork of this repo.

Issue #375, opened 2019-01-10
Record included twice in response when using included records.


shidel-dev:
This is an interesting case, and I am not 100% sure if this is a bug, but the way fast_jsonapi handles it is different than how AMS handles it. Also ember-data 2.13 seems to have issues with the way fast_jsonapi handles it as well.

Given a relationship like 
```
class Parent
  has_many :children
end

class Child
  belongs_to :parent
end
```

When a child is created and the include param is like `parent.children` the created child is both in the top level data field (as expected), and it is also in the included records.

With AMS it seems that the record in data is not repeated in the included records.

I could not find mention of this in the json api spec.

shidel-dev:
Are issues being actively

PR #381, opened 2019-01-21
records should be included in known_included_objects to avoid duplication

shidel-dev:
fixs #375


shidel-dev:
@shishirmk I curious to hear what is preventing this from being merged? Is anything missing that you would like to see? Handling this correctly seems like an obvious improvement. I believe the problem is clearly expressed in the test I wrote.

PR #384, opened 2019-01-26
WIP: make belongs_to polymorphic use `id_method_name`

### LINKED ISSUE #382: id_method_name not respected with polymorphic association
Without polymorphic association, I can do:

```rb
  belongs_to(
    :event,
    id_method_name: :event_uuid,
  )
```

It correctly uses `event_uuid` as the method to get the id of the association.

With polymorphic though:

```rb
  belongs_to(
    :resource,
    id_method_name: :resource_uuid,
    polymorphic: true
  )
```

`id_method_name` has no effect, it always uses `resource_id` as the id. I think this is a bug?

doits:
fixes #382

This is more or less a WIP to see what you guys say about this.

Some remarks:

- There could be a short cut to *not* fetch the associated record in case of `belongs_to` *without* a object block. In this specific case it might be enough to use the `xxx_type` column and `&:id_method_name` on the parent. This would remove the need to instantiate the associated record in that case. Though then the `fetch_id` method could not be used and the `belongs_to` case would be a completely separated case.
- `run_key_transform` is not yet cached (like in other methods)

What do you say about this in general?

I could think of a second way to make `belongs_to` have a complete different logic, which might speed `belongs_to` up though (if there is no need to fetch the association), but not sure if this is the way to go or if this makes the class too complex and if the `xxx_type` column is enough to get the class (or if there are some edge cases where this does not work).

### DIFF
--- a/spec/lib/object_serializer_polymorphic_spec.rb
+++ b/spec/lib/object_serializer_polymorphic_spec.rb
@@ -86,7 +86,7 @@ describe FastJsonapi::ObjectSerializer do
       expect(owner_relationship).to be_nil
     end
 
-    it 'should return nil for belongs_to when association is nil' do
+    it 'should return correct data for belongs_to with object_block when id_method_name is specified' do
       list = List.new
       list.id = 1
       list_hash = ListSerializer.new(list).to_hash


davidcelis:
This should probably also be updated to work with `has_one`

PR #396, opened 2019-03-05
Do not mutate options[:include] in ObjectSerializer#process_options

weilandia:
This pr makes it so `options[:include]` is not mutated when `ObjectSerializer` calls `process_options`.  This allows for the sharing of default include arrays that are stored in constants as frozen arrays.

For example:

``` ruby
class SomeSerializer
  DEFAULT_INCLUDE = [:movies, :'movies.advertising_campaign'].freeze
end

---
class SomeController
  def show
    render json: SomeSerializer.new(
        organization_rfq,
        include: SomeSerializer::DEFAULT_INCLUDE
      ).serialized_json
  end
end
```

Prior to these change this was not passible as you would get a `FrozenError: can't modify frozen Array` exception when [delete_if](https://github.com/Netflix/fast_jsonapi/blob/master/lib/fast_jsonapi/object_serializer.rb#L89) is called on `options[:include]`.

### DIFF
--- a/lib/fast_jsonapi/object_serializer.rb
+++ b/lib/fast_jsonapi/object_serializer.rb
@@ -86,7 +86,7 @@ module FastJsonapi
       raise ArgumentError.new("`params` option passed to serializer must be a hash") unless @params.is_a?(Hash)
 
       if options[:include].present?
-        @includes = options[:include].delete_if(&:blank?).map(&:to_sym)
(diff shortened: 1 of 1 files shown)


weilandia:
whoops, closing this in favor of https://github.com/Netflix/fast_jsonapi/pull/366

PR #402, opened 2019-03-21
Highlight that this is for JSON:API spec only

jessieay:
* The link was already there but I skipped over it on my first read.
This update makes the fact more prominent.
* I was testing to see if we wanted to move from AM Serializers to
fast_jsonapi but our API is not written according to the JSON:API spec
so, after converting one serializer over, I learned that this would not work for me.
* This update might save someone in my position the ~30 mins or so it
takes to bundle and write a serializer in the future. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,10 @@
 
 A lightning fast [JSON:API](http://jsonapi.org/) serializer for Ruby Objects.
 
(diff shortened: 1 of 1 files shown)


jessieay:
closing in favor of https://github.com/Netflix/fast_jsonapi/pull/403

Issue #393, opened 2019-02-25
Support for heterogeneous records while serializing


vivektiwary:
duplicate of https://github.com/Netflix/fast_jsonapi/issues/225

class Song < ActiveRecord::Base; end
class Pop < Song; end
class Rock < Song; end

class SongSerializer
  include FastJsonapi::ObjectSerializer
  attribute :title
end
class PopSerializer < SongSerializer
  attribute :description
end
class RockSerializer < SongSerializer
  attribute :duration
end

dvandersluis:
I am in need of this feature as well, is there any way to help make it happen?

Issue #225, opened 2018-05-21
Lookup for STI models


stelmashchuk:
First of all thanks for the great gem)
I use Single Table Inheritance in my project. 

```ruby
class Song < ActiveRecord::Base; end
class Pop < Song; end
class Rock < Song; end

class SongSerializer
  include FastJsonapi::ObjectSerializer
  attribute :title
end
class PopSerializer < SongSerializer
  attribute :description
end
class RockSerializer < SongSerializer
  attribute :duration
end
```
```ruby
@songs = [PopModel, RockModel]
render json: @songs
```

AMS makes automatic lookup and returns JSON with items with different attributes(```title, description``` for ```PopModel```; ```title, duration``` for ```RockModel```). 
Do you plan to add functionality like this? Thank you.

shishirmk:
@stelmashchuk we haven't worked on making the serializer work with heterogenous lists of models like the one you described above. It's not something we are looking into at the moment

stelmashchuk:
Thanks for answer

vivektiwary:
@shishirmk Is there any progress in this, are you guys thinking about giving this functionality?

top4ek:
Same question. Need to serialize a bunch of STI records with different attributes inherited.

PR #410, opened 2019-04-26
Handle a heterogeneous collection via serializers option

dvandersluis:
Proposal for allowing for a heterogeneous collection (or a single item of unknown type) to be serialized in a single response.

I settled on injecting the serializers to use in the call rather than having more metadata inside the serializer itself. I think this makes it clearer what the intent is, and also doesn't pollute the serializer itself.

```ruby
VehicleSerializer.new(
  [car, bus],
  serializers: { Car: CarSerializer, Bus: BusSerializer }
)
```

I foresee using this with `ApplicationSerializer` or the like as a base class (this makes more semantic sense to me). I experimented a bit with adding a generic abstract serializer to use, but I don't know if that's appropriate for this change. Another way to do this that I like would be to have a static method on `ObjectSerializer`, such as: 

```ruby
FastJsonapi::ObjectSerializer.serialize(
  [car, bus],
  serializers: { Car: CarSerializer, Bus: BusSerializer }
)
```

but to do so we would need a base class to instantiate.

I welcome any feedback, especially in terms of the interface.

References #225, #393.

### DIFF
--- a/lib/fast_jsonapi/object_serializer.rb
+++ b/lib/fast_jsonapi/object_serializer.rb
@@ -43,8 +43,9 @@ module FastJsonapi
 
       return serializable_hash unless @resource
 
-      serializable_hash[:data] = self.class.record_hash(@resource, @fieldsets[self.class.record_type.to_sym], @params)
-      serializable_hash[:included] = self.class.get_included_records(@resource, @includes, @known_included_objects, @fieldsets, @params) if @includes.present?
+      serializer_class = serializer_for(@resource)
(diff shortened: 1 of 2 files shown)


rmachielse:
Maybe it would be even cleaner to support `polymorphic: true`, just the way it works with relationships. That way you would get:

```ruby
VehicleSerializer.new([car, bus], polymorphic: true)
```

cazacutudor:
Hello @dvandersluis, we love to have this feature in place :heartbeat: . As you may already know, most of us have moved to https://github.com/jsonapi-serializer/jsonapi-serializer, if you don't mind, do you want to continue this pull-request there?

PR #424, opened 2019-06-07
allow multiple meta blocks and merge them if present

doits:
Useful for includes and inheritance:

```rb
module SomeModule
  extend ActiveSupport::Concern

  included do
    meta do
      {
        module_meta: true
      }
    end
  end
end

class UserSerializer
  include SomeModule

  meta do
    {
      user_meta: true
    }
  end
end

class EmployeeSerializer < UserSerializer
  meta do
    {
      employee_meta: true
    }
  end
end

EmployeeSerializer.new(employee).serializable_hash

{
  data: {
    [...]
    meta: {
      module_meta: true,
      user_meta: true,
      employee_meta: true
    }
  }
}
```


doits:
One thing to discuss is whether `deep_merge` or `merge` should be used. Only makes a difference in the exotic case of different meta blocks returning same keys though.

Otherwise I think this change makes composing serializers out of different modules and/or inheritance much more simpler, because each module or class can have meta blocks.

PR #425, opened 2019-06-10
Include `data` key when lazy-loaded relationships are included

### LINKED ISSUE #357: Exclude data from relationships (unless included)
This is re-opening https://github.com/Netflix/fast_jsonapi/issues/252, because AFAIK fast_jsonapi does not support this in the current version.

In our scenario it is costly to fetch the ids of related resources, so not returning "data", but just "links" would great. The goal is to return something like this - Return a link (URL) to the related records (this is solved), but don't return the ids in data (this is AFAIK not solved):

```
"attributes": {
  ...
},
"relationships": {
  "comments": {
    "links": {
      "related": "http://localhost/api/v1/users/1/comments"
    }
},
```

[jsonapi-serializers](https://github.com/fotinakis/jsonapi-serializers) does this via [`has_many :comments, include_data: true`](https://github.com/fotinakis/jsonapi-serializers#control-links-and-data-in-relationships).

dpikt:
Resolves #357 

When a relationship is listed under `include:`, it will add the data key to the record hash, allowing the included relationships to be reconstructed.

### DIFF
--- a/lib/fast_jsonapi/object_serializer.rb
+++ b/lib/fast_jsonapi/object_serializer.rb
@@ -43,7 +43,7 @@ module FastJsonapi
 
       return serializable_hash unless @resource
 
-      serializable_hash[:data] = self.class.record_hash(@resource, @fieldsets[self.class.record_type.to_sym], @params)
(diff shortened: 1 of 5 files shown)


kln:
up

PR #442, opened 2019-09-13
Allow "if: Proc..." on link

listrophy:
Just like `attribute`, this PR allows the dev to specify links as follows:

```ruby
link(:download, if: ->(obj, params) { params && params[:signed_url] }) do |obj, params|
  params[:signed_url]
end
```


jopotts:
Looks like an improvement. You're welcome to add the PR here too: https://github.com/fast-jsonapi/fast_jsonapi/issues/1

PR #446, opened 2019-10-29
fix id method bugs

aub:
For polymorphic has_many relations, instead of getting the ids in a batch, the code loops over each object to get its type and id. In Relationship#id_hash_from_record it was hardcoded to get the id by calling record.id, which overrides the id_method_name options passed into the relationship.

I tried fixing this by only changing the code in Relationship to dynamically call the id_method_name on the object, but unfortunately on has_many relationships the default value for id_method_name would be passed as if the ids would be pulled in a batch (for example groupee_ids). That default works for regular has_manys where we get the ids all at once but fails for polymorphic ones where it loops over each object.

So, fixing the problem required passing the polymorphic information to compute_id_method_name, which now defaults to id rather than id_method_name_from_relationship in the case where the relationship is polymorphic.

This also fixes a bug where the record_id is not carried over through inheritance.

### DIFF
--- a/lib/fast_jsonapi/object_serializer.rb
+++ b/lib/fast_jsonapi/object_serializer.rb
@@ -124,6 +124,7 @@ module FastJsonapi
         subclass.cached = cached
         subclass.set_type(subclass.reflected_record_type) if subclass.reflected_record_type
         subclass.meta_to_serialize = meta_to_serialize
+        subclass.record_id = record_id
       end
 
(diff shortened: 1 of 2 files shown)


kapilnarula:
It seems Netflix has abandoned this project. The community created a new
fork to continue supporting this project. Please refer

https://github.com/fast-jsonapi/fast_jsonapi

PR #450, opened 2019-12-18
Fix Travis CI build by allowing newer Bundler versions

attilahorvath:
The Travis CI build currently fails because it's trying to use a newer version of Bundler which is disallowed by the `fast_jsonapi.gemspec` file.

This is an attempt to fix the builds by allowing for these newer Bundlers.


attilahorvath:
Looks like there are other issues with this and it's fixed anyway in the [fork](https://github.com/fast-jsonapi/fast_jsonapi) so I'll close this.

PR #458, opened 2020-03-28
Update development dependencies

alejandroperea:
This PR updates some development dependencies to its latest version.

Also requires ActiveSupport >=5.2, as ActiveSupport 4 has arrived at its EOL.


alejandroperea:
Project seems to be abandoned. Switching to community one.

Issue #459, opened 2020-04-02
Unexpected behavior when object belongs_to :size


gkpacker:
Hello!

I'm trying to serialize my product object but it responds to `size`, so it is treated as a collection in:
```rb
    def is_collection?(resource, force_is_collection = nil)
      return force_is_collection unless force_is_collection.nil?

      resource.respond_to?(:size) && !resource.respond_to?(:each_pair)
    end
```
and ends up breaking the code:

```
[1] pry(#<Api::V1::ProductsController>)> ProductSerializer.new(@product).serializable_hash
NoMethodError: undefined method `each' for #<Product:0x00007f8f563881b8>
```

It is intended to be that way?

Workaround:
```
ProductSerializer.new(@product, is_collection: false).serializable_hash
```

Thanks!

stas:
This is now merged in the forked project!

gkpacker:
Nice, I'll close the issue!

PR #460, opened 2020-04-11
Check #each instead of #size for is_collection?

marksiemers:
See if a resource responds to 'each' instead of 'size' as
things other than collections can respond to 'size' (e.g. File instance)


marksiemers:
Should address this issue: https://github.com/Netflix/fast_jsonapi/issues/459

